PR #5, opened 2019-04-30
build-llvm.py: Split into a two-stage build

nathanchance:
This will make builds more reproducible and easier to manage over time,
as we don't have to worry as much about the user's environment because
they just need to be able to build stage 1 then stage 2 will be
consistent across environments.

This does take longer so an option '--stage1-only' is introduced for
users who have less powerful systems and instances like bisecting where
we just want a compiler to test something really quick.

There is another option introduced, '--thin-lto', which enabled ThinLTO
because we are using ld.lld. This can save some cycles (~5% speed up in
my experience) when using the toolchain.

The cmake defines are now logically split up into a couple of different
functions for easy ingestion and decisions based on what stage we are
at.


nathanchance:
Note: I am going to be stress testing this over the next few hours, I'll push any follow up fixes separately.

PR #10, opened 2019-05-01
build-binutils.py: Rename root to root_folder for consistency

nathanchance:
This is the result of:

```
$ sed -i 's/root/root_folder/g' build-binutils.py && yapf build-binutils.py
```

I would prefer that #7 and #8 are merged ahead of this because I can run that one liner upon rebase super easily. I've just pushed it now as a heads up that I want this done.


nathanchance:
This is ready to go in now.

PR #13, opened 2019-06-04
Select LLVM_ENABLE_PLUGINS

nathanchance:


nathanchance:
I did that in b6dc15b55bf9f8e3968b19a922b6b448695007d8. Is that message good enough or should I make it clearer?

stephenhines:
Ah, that's fine. Thanks for fixing this up.

nathanchance:
Thanks for the review!

PR #14, opened 2019-06-04
build-llvm.py: Disable some more compiler-rt options

nathanchance:
We do not use these things, which will save some build targets on stage
2. This was inspired by https://crrev.com/c/1629950.

Number of stage 2 targets before this commit: 4031

Number of stage 2 targets after this commit: 3703


nathanchance:
Thanks for the review!

PR #16, opened 2019-06-16
PGO fixes

nathanchance:
A few fallout fixes from #15 after harder testing, the changes should be self explanatory.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -639,10 +639,12 @@ def generate_pgo_profiles(args, dirs):
     utils.print_header("Building PGO profiles")
 
     # Run kernel/build.sh
-    subprocess.run(
-        [dirs.root_folder.joinpath("kernel", "build.sh"), '-t', args.targets],
-        check=True,
-        cwd=dirs.build_folder.as_posix())
+    subprocess.run([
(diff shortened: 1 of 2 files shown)


build-llvm.py
```diff
@@ -639,10 +639,12 @@ def generate_pgo_profiles(args, dirs):
...
 
     # Run kernel/build.sh
-    subprocess.run(
-        [dirs.root_folder.joinpath("kernel", "build.sh"), '-t', args.targets],
-        check=True,
-        cwd=dirs.build_folder.as_posix())
+    subprocess.run([
+        dirs.root_folder.joinpath("kernel", "build.sh"), '-b',
+        dirs.build_folder, '-t', args.targets
+    ],
```
ojeda: If you are separating the lines, maybe it looks better to put an option per line, i.e. something like:

    subprocess.run([
        dirs.root_folder.joinpath("kernel", "build.sh"),
        '-t', dirs.build_folder,
        '-b', args.targets
    ],

nathanchance: I agree that looks better but I've been formatting the code with [yapf](https://github.com/google/yapf) for consistency and it chose this look. It reformats your suggestion to the current look.

ojeda: Never mind then! :)

PR #19, opened 2019-06-26
build-llvm.py: Only apply native optimizations if specifically requested

nathanchance:
This causes issues with older versions of LLVM on some people's
hardware. Make it opt in it won't mess with people who just run
'./build-llvm.py' and people who want it can still enable it.

At the same time, drop the -O2 flag; release LLVM builds are compiled
with -O3 and it overrides our choice anyways (always appears after
CMAKE_C_FLAGS).

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -98,6 +98,13 @@ def parse_parameters(root_folder):
                         type=str,
                         default=os.path.join(root_folder.as_posix(),
                                              "install"))
+    parser.add_argument("--march-native",
+                        help=textwrap.dedent("""\
+                        Add -march=native and -mtune=native to CFLAGS to further optimize the toolchain for
+                        the host processor.
+
+                        """),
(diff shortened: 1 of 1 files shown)


build-llvm.py
```diff
@@ -104,6 +104,14 @@ def parse_parameters(root_folder):
                         By default, the script always updates the LLVM repo before building. This prevents
                         that, which can be helpful during something like bisecting.
 
+                        """),
+                        action="store_true")
+    parser.add_argument("-o",
```
nickdesaulniers: bikeshed on naming; `-o` is usually "what to call the output," so using it to enable an option violates the principle of least surprise.  Please just call the option `--march-native` or something of the like.

nathanchance: Sure

nathanchance: Done.

PR #27, opened 2019-07-11
kernel/build.sh: Improve handling of Linux source tarballs

dileks:
Some little improvements in handling Linux source tarballs:

Save download band width by using *.tar.xz tarballs from <kernel.org>
(Linux v5.2: approx. 50MiB).

Simplify the decompression of tarballs as modern tar versions are
smart enough to detect the (de)compressor.

Further improvements (TBD):
Verify the sign file [2]  of the desired Linux source tarball.

Details see tc-build issue #26:
"Linux sources: Download XZ tarball, simplify decompressing and verify sign-file"

[1] https://github.com/ClangBuiltLinux/tc-build/issues/26
[2] https://mirrors.edge.kernel.org/pub/linux/kernel/v5.x/linux-5.2.tar.sign

### DIFF
--- a/kernel/build.sh
+++ b/kernel/build.sh
@@ -47,7 +47,7 @@ if [[ -n ${SRC_FOLDER} ]]; then
     cd "${SRC_FOLDER}" || exit 1
 else
     LINUX=linux-5.1
-    LINUX_TARBALL=${TC_BLD}/kernel/${LINUX}.tar.gz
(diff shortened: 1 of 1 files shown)


ojeda:
> Verify the sign file [2] of the desired Linux source tarball.

Like in tc-build, it is easier to compute a hash and compare it with a hardcoded one. As long as whoever updates the code to a different version verifies the signature file before computing the hash that will be hardcoded in the source, of course.

PR #38, opened 2019-07-24
build-llvm.py: Add '--use-good-revision'

nathanchance:


dileks:
s/This is a good known revision/This is a known good revision

nathanchance:
I will merge this one last as it will have conflicts with #34 (the `Directories()` to `Directories`) and #37 (`no_pull` to `no_update`).

PR #40, opened 2019-07-26
build-llvm.py: Add '--assertions'

nathanchance:
There was an issue with https://github.com/llvm/llvm-project/commit/5354c83ece00690b4dbfa47925f8f5a8f33f1d9e that only appears when
assertions are enabled (because it only adds an assertion). Add a way to
turn them on in a release configuration so that compiles are still
relatively quick.


nathanchance:
I am going to respin this with the idea I had in #42.

PR #47, opened 2019-08-06
build-llvm.py: Improve '--check-targets' help text

nathanchance:
Make it explicitly clear what this flag does and how it is used.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -111,10 +111,12 @@ def parse_parameters(root_folder):
     parser.add_argument("--check-targets",
                         help=textwrap.dedent("""\
                         By default, no testing is run on the toolchain. If you would like to run unit/regression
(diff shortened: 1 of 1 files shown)


dileks:
Small improvement:
1. Add separate Example line (with a newline)
2. Typo s/paramater/parameter
3. Mention list is space separated?

```
The values passed to this paramater will be automatically concatenated with 'check-' (e.g.
'--check-targets clang llvm' becomes 'check-clang' and 'check-llvm').

The values passed to this parameter will be automatically concatenated with 'check-'  (space separated list).

Example: '--check-targets clang lld' will run tests for clang and lld
```

nathanchance:
> 1. Add separate Example line (with a newline)

Sure.

> 2. Typo s/paramater/parameter

Done, thanks.

> 3. Mention list is space separated?

I think the example covers this; if the user can't figure that out, we have bigger problems :)

Thanks for the review!

PR #50, opened 2019-09-17
Linux 5.3 and some assorted fixes

nathanchance:
Bump the Linux version used for PGO benchmarking and add some fixes for issues I noticed when doing this update.


dileks:
Thanks @nathanchance

```diff
-    [[ -f ${LINUX_PATCH} ]] && { git apply "${LINUX_PATCH}" || exit ${?}; }
+    [[ -f ${LINUX_PATCH} ]] && { patch -p1 < "${LINUX_PATCH}" || exit ${?}; }
```

You can pass `-p1` to `git apply` means `git apply -p1 $LINUX_PATCH`.

nathanchance:
@dileks thank you for the comment. The problem is not the stripping of the path, it is something else. With `git apply` currently, instead of saying `Patching <file>`, it says `Skipping <file>` and I cannot figure out why; if it were the path issue, it would say something about not being able to find the file to patch. Rather than mess around with `git apply`, going with `patch -p1` works just fine and avoids any issues.

PR #52, opened 2019-09-18
kernel/build.sh: Show make command using set -x and only show warnings/errors

nathanchance:
This will make it easier to audit build failures because the exact make
command can be copy and pasted from the terminal.


nathanchance:
This won't make things more noisy because the `-s` addition to `make` silences the regular print outs such as `CC <file>` and `LD <file>` in the kernel build, instead just showing warnings/errors (which it already does).

The `set -x` is placed down further so that all of the internal logic of downloading/extracting the tarball and such does not contribute to the script output.

nathanchance:
Thanks for the review! :)

PR #53, opened 2019-09-21
build-binutils.py: Don't optimize for native by default

wloot:


nathanchance:
Tentatively LGTM but I will test when I get home.

Issue #18, opened 2019-06-26
Backend crash on Debian stretch


nathanchance:
Thomas Gleixner reported that running `./build-llvm.py` caused the build to explode.

https://lore.kernel.org/lkml/alpine.DEB.2.21.1906261711540.32342@nanos.tec.linutronix.de/

Log on gist: https://gist.github.com/b3be9ce876866c15bbce240328035bcb

My guess is this is related to `-march=native` because the bug happens during lowering. I am having him try my opt in patch now, which I'll push for review once I get confirmation that it works: https://github.com/nathanchance/tc-build/commit/9f1ae41cd4246f9e4d011542f094aa0df2c069b4

PR #55, opened 2019-09-21
build-llvm.py: Always optimize for native at intermediate build

wloot:


nathanchance:
1. There needs to be a check for stage 1 (because otherwise we will potentially run into #18 again).

2. I’m not convinced this is useful. `-march=native` won’t necessarily produce faster code. Some benchmarks showing a real world improvement would be kind of nice (`hyperfine` is a good tool).

wloot:
> 1. There needs to be a check for stage 1 (because otherwise we will potentially run into #18 again).
> 2. I’m not convinced this is useful. `-march=native` won’t necessarily produce faster code. Some benchmarks showing a real world improvement would be kind of nice (`hyperfine` is a good tool).

`not stage == get_final_stage(args)` will be false when build_stage1_only enabled, as get_final_stage() returns 1 at that time.

nathanchance:
You haven't considered the default case of the script, which is a two stage build. Thus, `not stage == get_final_stage(args)` will evaluate to `not stage == 2`, which means that stage 1 will be built with `-march=native`, which was the root cause of #18.

There are three different stage combinations:

1. stage 1 only. This is already handled fine by our `-march` flag.
2. stage 1 and stage 2. This is the default of the script. We cannot apply `-march=native` automatically because of #18.
3. stage 1, stage 2, and stage 3. This is used when doing PGO. We could automatically apply `-march=native` to stage 2 (the instrumented compiler) because the stage 1 compiler will be able to handle that without issues.

Based on the above, `if instrumented_stage(args, stage)` would work for the third point but I really don't think this is worthwhile. [`hyperfine`](https://github.com/sharkdp/hyperfine) might be able to tell but the kernel build varies in time too much to be able to pin down for sure if adding `-march=native` would add any sort of speed up.

wloot:
> You haven't considered the default case of the script, which is a two stage build. Thus, `not stage == get_final_stage(args)` will evaluate to `not stage == 2`, which means that stage 1 will be built with `-march=native`, which was the root cause of #18.
> 
> There are three different stage combinations:
> 
> 1. stage 1 only. This is already handled fine by our `-march` flag.
> 2. stage 1 and stage 2. This is the default of the script. We cannot apply `-march=native` automatically because of #18.
> 3. stage 1, stage 2, and stage 3. This is used when doing PGO. We could automatically apply `-march=native` to stage 2 (the instrumented compiler) because the stage 1 compiler will be able to handle that without issues.
> 
> Based on the above, `if instrumented_stage(args, stage)` would work for the third point but I really don't think this is worthwhile. [`hyperfine`](https://github.com/sharkdp/hyperfine) might be able to tell but the kernel build varies in time too much to be able to pin down for sure if adding `-march=native` would add any sort of speed up.

Means we can't apply -march=native on stage1 of default mode? I am still confused that the final build will have no -march=native flag be setted by default.

nathanchance:
Correct, because the compiler for stage 1 is the host compiler, which might not properly support `-march=native`.

nathanchance:
I decided to benchmark if `--march=native` made any difference in LLVM and kernel compile time with the script as of 439c648a4f7051ec6360b96ed44ae6494474158a (i.e., without this change) with [`hyperfine`](https://github.com/sharkdp/hyperfine) (see my script [here](https://gist.github.com/nathanchance/b12be3e82f0b038f8bd442c9e3ec9eb3)) and here are the results from three different servers/processors (which I found surprising). There were 30 runs done for each compiler build and 60 runs done for each kernel build (x86_64 defconfig on Linux 5.3).

## [`c2.medium.x86`](https://www.packet.com/cloud/servers/c2-medium-epyc/):

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./build-llvm.py` | 802.050 ± 2.071 | 795.101 | 804.847 | 1.00 |
| `./build-llvm.py --march=native` | 811.412 ± 1.890 | 808.501 | 815.833 | 1.01 ± 0.00 |

### Kernel build times:

| Compiler | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Clang compiled without `-march=native` | 84.979 ± 0.489 | 83.526 | 85.595 | 1.00 |
| Clang compiled with `-march=native` | 85.137 ± 0.438 | 83.525 | 85.602 | 1.00 ± 0.01 |

## [`m1.xlarge.x86`](https://www.packet.com/cloud/servers/m1-xlarge/):

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./build-llvm.py` | 766.981 ± 1.000 | 765.095 | 769.026 | 1.00 |
| `./build-llvm.py --march=native` | 775.329 ± 0.948 | 773.021 | 777.543 | 1.01 ± 0.00 |

### Kernel build times:

| Compiler | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Clang compiled without `-march=native` | 90.125 ± 0.516 | 88.743 | 91.343 | 1.00 |
| Clang compiled with `-march=native` | 90.216 ± 0.606 | 89.136 | 91.678 | 1.00 ± 0.01 |

## [`m2.xlarge.x86`](https://www.packet.com/cloud/servers/m2-xlarge/):

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./build-llvm.py` | 595.829 ± 0.374 | 595.219 | 596.573 | 1.00 |
| `./build-llvm.py --march=native` | 608.040 ± 0.374 | 607.371 | 609.041 | 1.02 ± 0.00 |

### Kernel build times:

| Compiler | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Clang compiled without `-march=native` | 74.878 ± 0.248 | 74.275 | 75.412 | 1.00 |
| Clang compiled with `-march=native` | 84.027 ± 0.252 | 83.476 | 84.470 | 1.12 ± 0.01 |

I don't think that this change is worthwhile and I am kind of considering ripping out `--march=` because it never appears to help... Please reopen if you feel otherwise and can justify it with real numbers.

nickdesaulniers:
build times go up? 🥚 on 🤕

nathanchance:
I will say that @kerneltoast and @kdrag0n pointed out that the massive regression in build times on the m2.xlarge.x86 is a result of AVX512 instructions, which are known to cause the CPU to downclock: https://reviews.llvm.org/D67259

That change should land in LLVM 10, although I am not sure if that affects `-march=native`; I will do another set of benchmarks with current tip of tree on that CPU and see.

At any rate though, `-march=native` never seems to beat no `-march=native` on any of the three machines or benchmarks that I tested.

nickdesaulniers:
This is useful to talk about at the meetup.  IIRC, Google internal engineers had the same findings in regard to avx512.

ojeda:
I guess compiling kernels and compilers isn't fp-heavy enough to offset the downclock.

nickdesaulniers:
@nathanchance , what about `-march=native -mno-avx512` (assuming clang even recognizes that flag)? Presumably you'd still want to make use of other ISA extensions?

kdrag0n:
I retested it on a [`g2.large.x86`](https://www.packet.com/cloud/servers/g2-large/) machine (Xeon Gold 6126 CPU) with LLVM 10 commit https://github.com/llvm/llvm-project/commit/e64f7bfef using @nathanchance's script and found similar results:

### LLVM build times
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./build-llvm.py` | 672.461 ± 0.768 | 671.444 | 674.163 | 1.0 |
| `./build-llvm.py --march=native` | 697.723 ± 0.539 | 696.412 | 698.611 | 1.0 |

### Kernel build times
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Clang compiled without `-march=native` | 71.293 ± 0.146 | 71.011 | 71.677 | 1.0 |
| Clang compiled with `-march=native` | 71.406 ± 0.136 | 71.009 | 71.701 | 1.0 |

LLVM builds with `-march=native` are several seconds longer (4% of overall build time), while kernel build times are nearly identical if not slightly slower. Note that these tests were done on LLVM 10 tip so the AVX512 width change should be included. I'm in favor of ditching the flag based on these results, but I'm wondering if Polly would show an improvement due to its more complicated math.

nathanchance:
I am currently running a set of benchmarks with LLVM 9.0.0 on a few different servers around `-march=native` vs. `-mtune=native`. Depending on how the results turn out, we can either switch to only `-mtune` or just get rid of it all together if it does not help.

nathanchance:
Final results from my testing (script available [here](https://github.com/nathanchance/tc-build/blob/0f62ecbd5806eb9bb50fd5350f28079727ea6ab5/benchmark/march-mtune-native.sh)):

## [`c2.medium.x86`](https://www.packet.com/cloud/servers/c2-medium-epyc/)

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 804.806 ± 1.718 | 802.612 | 809.017 | 1.00 |
| `-m{arch,tune}=native` | 814.567 ± 1.656 | 812.166 | 817.801 | 1.01 ± 0.00 |
| `-mtune=native` | 804.893 ± 1.721 | 803.011 | 809.654 | 1.00 ± 0.00 |

### Kernel build times:

Building `x86_64_defconfig` on Linux 5.3 using `clang` with:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 85.462 ± 0.353 | 83.970 | 85.829 | 1.00 ± 0.01 |
| `-m{arch,tune}=native` | 85.303 ± 0.466 | 83.768 | 85.811 | 1.00 |
| `-mtune=native` | 85.525 ± 0.393 | 84.192 | 85.941 | 1.00 ± 0.01 |

## [`x2.xlarge.x86`](https://www.packet.com/cloud/servers/x2-xlarge/)

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 531.691 ± 0.508 | 530.661 | 532.694 | 1.00 ± 0.00 |
| `-m{arch,tune}=native` | 582.782 ± 0.495 | 581.929 | 583.752 | 1.10 ± 0.00 |
| `-mtune=native` | 531.640 ± 0.373 | 530.858 | 532.317 | 1.00 |

### Kernel build times:

Building `x86_64_defconfig` on Linux 5.3 using `clang` with:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 61.269 ± 0.200 | 60.435 | 61.545 | 1.00 ± 0.00 |
|`-m{arch,tune}=native` | 68.898 ± 0.256 | 67.788 | 69.310 | 1.13 ± 0.01 |
| `-mtune=native` | 61.183 ± 0.167 | 60.438 | 61.426 | 1.00 |

## [`g2.large.x86`](https://www.packet.com/cloud/servers/g2-large/)

### LLVM build times:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 561.676 ± 0.998 | 560.382 | 563.735 | 1.00 ± 0.00 |
| `-m{arch,tune}=native` | 585.565 ± 0.851 | 584.119 | 587.804 | 1.04 ± 0.00 |
| `-mtune=native` | 561.343 ± 0.742 | 560.279 | 562.884 | 1.00 |

### Kernel build times:

Building `x86_64_defconfig` on Linux 5.3 using `clang` with:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| No native tuning | 65.265 ± 0.128 | 64.923 | 65.520 | 1.00 |
| `-m{arch,tune}=native` | 67.476 ± 0.163 | 67.020 | 67.822 | 1.03 ± 0.00 |
| `-mtune=native` | 65.286 ± 0.172 | 64.697 | 65.654 | 1.00 ± 0.00 |

## Conclusion

When AVX512 is involved, there is a noticeable degradation in performance; when the CPU does not support, `-march=native` makes no discernible difference (most likely because of what @ojeda mentioned, our workload is not floating point heavy so these optimizations don't do anything for us).

I think `--march=` should be nuked from the script to discourage users from enabling it but adding something like  `--cflags` so that it is easy to add flags like these to test in the future: https://github.com/nathanchance/tc-build/commit/29f643a781413c2be5fd3481727235450e747b45

ojeda:
Yeah, AVX-512 makes things counter-intuitive on Intel. It could be argued that GCC/Clang/etc. shouldn't enable all instruction sets for a given model if they depend on the workload from `native` (since most developers map `native` to "whatever is best for this chip"); using a similar argument as they do for `O2` vs. `O3`.

PR #56, opened 2019-09-22
README.md: Add flex and bison to needed packages

wloot:
as binutils build request it, otherwise the build fails:
binutils/missing: 81: binutils/missing: flex: not found
WARNING: 'flex' is missing on your system.
         You should only need it if you modified a '.l' file.
         You may want to install the Fast Lexical Analyzer package:
         <http://flex.sourceforge.net/>
make[2]: *** [Makefile:1108: syslex.c] Error 127

binutils/missing: 81: binutils/missing: bison: not found
WARNING: 'bison' is missing on your system.
         You should only need it if you modified a '.y' file.
         You may want to install the GNU Bison package:
         <http://www.gnu.org/software/bison/>
make[2]: *** [Makefile:1111: sysinfo.c] Error 127

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ These scripts have been tested in a Docker image of the following distributions,
 * ### Debian/Ubuntu
 
   ```
-  apt install ca-certificates ccache clang cmake curl file gcc g++ git make ninja-build python3 texinfo zlib1g-dev bison flex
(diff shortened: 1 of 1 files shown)


nathanchance:
Can you please keep the packages alphabetized?

PR #58, opened 2019-09-23
Fix '--use-good-revision' while improving compatibility with older versions of git

nathanchance:
Each commit message should be fairly self explanatory :)


nathanchance:
Thanks for the review :)

PR #62, opened 2019-11-26
tc-build: Update Linux to 5.4 and bump the known good revision

nathanchance:
Verified on a fresh checkout with:

```bash
$ ./build-binutils.py && \
./build-llvm.py --check-targets clang lld llvm --pgo --use-good-revision && \
PATH_OVERRIDE=${PWD}/install/bin ./kernel/build.sh --allyesconfig
```


nathanchance:
Thanks all!

PR #63, opened 2020-01-27
tc-build: Update Linux to 5.5 and bump the known good revision

nathanchance:


wloot:
Works here as well
***
	--projects "clang;compiler-rt;libunwind;lld;polly" \
	--targets "ARM;AArch64;X86" \
	--build-stage1-only \
	--install-stage1-only
***
and success to build k5.5 and k4.19

nathanchance:
Thanks for the review!

PR #65, opened 2020-02-07
build-llvm.py: Look for clang-11 when building stage 1

nathanchance:
Also, add a variable so this becomes easier to bump. Eventually, I would like to do this similar to `driver.sh` in the CI repo where we just grab the latest version from the LLVM repo but that becomes hard in Python...

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -267,17 +267,22 @@ def versioned_binaries(binary_name):
     :return: List of versioned binaries
     """
 
-    # There might be clang-6 to clang-11
+    # There might be clang-7 to clang-11
     tot_llvm_ver = 11
     try:
         response = request.urlopen(
             'https://raw.githubusercontent.com/llvm/llvm-project/master/llvm/CMakeLists.txt'
         )
-        data = response.read().decode('utf-8')
-        tot_llvm_ver = re.search(r'^.*LLVM_VERSION_MAJOR\ (\d{2})\)', data).group(0)
+        to_parse = None
+        data = response.readlines()
+        for line in data:
+            line = line.decode('utf-8').strip()
+            if "set(LLVM_VERSION_MAJOR" in line:
+                to_parse = line
+        tot_llvm_ver = re.search('\d+', to_parse).group(0)
     except URLError:
         pass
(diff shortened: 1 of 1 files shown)


msfjarvis:
> Also, add a variable so this becomes easier to bump. Eventually, I would like to do this similar to `driver.sh` in the CI repo where we just grab the latest version from the LLVM repo but that becomes hard in Python...

[Holy mother of Bash hacks](https://github.com/ClangBuiltLinux/continuous-integration/blob/999014f5d3a48231833a2aa22c79ca2711472d9c/driver.sh#L183)... I'll see if I can port this somehow.

nathanchance:
> [Holy mother of Bash hacks](https://github.com/ClangBuiltLinux/continuous-integration/blob/999014f5d3a48231833a2aa22c79ca2711472d9c/driver.sh#L183)...

Don’t hate on my bash :(

> I'll see if I can port this somehow.

But thank you, a good Python solution would be great so we can avoid manually bumping :)

msfjarvis:
Maybe we should just use the local copy of the file?

nathanchance:
No, if the script is being run for the first time, we won’t have the repo.

The main issue I foresee with this is the script won’t work without internet now. Before this, if `--no-update` were passed, there would be no need for internet but now this requires it. Not sure how much that actually matters and in reality, we could just have the fallback if we can’t reach the URL.

msfjarvis:
Yeah that's what I thought. I'll add a fallback to clang 11.

nathanchance:
Thanks :)

msfjarvis:
Well that took a while. Confirmed the test cases that came to mind

- [x] Internet connected, LLVM 11 snapshot installed, tc-build uses {clang,lld}-11
- [x] Internet disconnected, LLVM 11 snapshot installed, tc-build uses {clang,lld}-11
- [x] Internet disconnected, LLVM 11 snapshot installed, tc-build master uses {clang,lld}-10
- [x] Internet connected, LLVM 11 snapshot installed, tc-build master uses {clang,lld}-10

msfjarvis:
image

PR #67, opened 2020-02-08
tc-build: Update binutils to 2.34

nathanchance:
Fix the comment after commit 7fc8133 ("utils: Update to binutils 2.33.1
and other small improvements") where we switched to xz compression.


nathanchance:
Thanks!

PR #68, opened 2020-02-08
build-llvm.py: Only enable assertions on the final build

nathanchance:
We do not want assertions to be enabled on a stage 2 PGO build otherwise
profile generation will take ages.


nathanchance:
Thanks for the review!

PR #69, opened 2020-02-11
build-llvm.py: Add an option to perform shallow clones

kdrag0n:
There are times when one-time builds are desired (e.g. CI runs). Cloning the LLVM repo with history is wasteful in such cases because the history will never be used.

Add an option to perform a shallow clone rather than a full clone to speed up builds when history and updating are not required.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -195,6 +195,15 @@ def parse_parameters(root_folder):
 
                         """),
                         action="store_true")
+    parser.add_argument("-s",
+                        "--shallow-clone",
+                        help=textwrap.dedent("""\
+                        Only fetch the required objects and omit history when cloning the LLVM repo. This
+                        speeds up the initial clone, but may break updating to later revisions and thus
+                        necessitate a re-clone in the future.
+
+                        """),
+                        action="store_true")
     parser.add_argument("-p",
                         "--projects",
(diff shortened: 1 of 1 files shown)


nathanchance:
Maybe the help text should have stronger wording that this is an option that should only be used for one time instances like CI? As I see it now, a user might use `-s` to save on the initial step but run into issues trying to run the script again because the update path does not handle `-s` (meaning the script will explode in that instance).

We could either make `-s` work in the update path (either remove and reclone or just `git fetch --depth=1` + `git reset --hard FETCH_HEAD` like we do in [continuous-integration](https://github.com/ClangBuiltLinux/continuous-integration/blob/4e86683b649b5d06574aaed63bda221a894e7b86/driver.sh#L310-L311)) or just clarify in the help text and leave it up to the user.

It might even be worthwhile to just add a `--ci` option or something that tries to optimize paths like these if you can think of anymore that might be helpful.

kdrag0n:
Updating works fine after a shallow clone. Git is still able to figure out what changed and pull a small delta accordingly without having to download the full history.

The only other option I can think of that would benefit CIs is `--no-ccache`, but I don't see a reason to add a CI-specific metaflag just to enable two other existing options, especially since argparse doesn't have native support for metaflags.

PR #70, opened 2020-02-11
readme: Add missing depdencies to package install lists

kdrag0n:
During PGO kernel builds, OpenSSL (for certificate generation) and libelf (for the x86 ORC unwinder) are used. Add these packages as dependencies in the distro package install lists to fix build errors on freshly provisioned machines.

Tested on a fresh install of Ubuntu 18.04.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ These scripts have been tested in a Docker image of the following distributions,
 * ### Debian/Ubuntu
 
   ```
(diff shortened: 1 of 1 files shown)


nathanchance:
I think the README needs to be worded better around PGO because I have a note below these packages... Let me whip something up real quick.

nathanchance:
@kdrag0n what do you think of https://github.com/nathanchance/tc-build/commit/2c9e55994df0e99e5fae92bfa56d871862cd3315?

kdrag0n:
That looks better, thanks.

PR #71, opened 2020-02-11
README: Hoist PGO packages into list of packages to install

nathanchance:
Originally, I had listed the PGO packages as an aside because they were
just for building the kernels and I did not want to bloat people's
systems.

However, it has caused some headaches over time because they did not
realize that they needed more packages than the ones listed because the
PGO packages were listed in a paragraph below and it was not very clear.

Fix this by just moving the packages needed for --pgo to work into the
main list of packages; it won't hurt people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,26 @@ These scripts have been tested in a Docker image of the following distributions,
 * ### Debian/Ubuntu
 
   ```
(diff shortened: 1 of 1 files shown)


nathanchance:
Done, thanks!

PR #73, opened 2020-03-15
Fix PGO builds when targeting x86

kdrag0n:
Stable 5.5.9 includes a [fix](https://lore.kernel.org/lkml/20200310123644.028693381@linuxfoundation.org/) for the following error, which breaks PGO builds targeting x86 on newer LLVM revisions:

```
ld.lld: error: duplicate symbol: __force_order
>>> defined at kaslr_64.c
>>>            arch/x86/boot/compressed/kaslr_64.o:(__force_order)
>>> defined at pgtable_64.c
>>>            arch/x86/boot/compressed/pgtable_64.o:(.bss+0x2000)
```


nathanchance:
Confirmed this works, thanks!

PR #74, opened 2020-03-20
build-llvm.py: Enable Polly by default

nathanchance:
More and more Android kernel developers are starting to enable Polly.
This was recently enabled in AOSP's LLVM builds:

https://android.googlesource.com/toolchain/llvm_android/+/6ba474f9d931f696860cec49b8c1dd62ac1e5a46

Do the same thing in our build. It only adds approximately 150 targets
to the final LLVM stage (comparing --build-stage1-only with and without
this patch, 3714 -> 3870).

While we're at it, alphabetize the projects.

cc @kdrag0n @0ctobot @lazerl0rd

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -220,7 +220,7 @@ def parse_parameters(root_folder):
 
                         """),
                         type=str,
-                        default="clang;lld;compiler-rt")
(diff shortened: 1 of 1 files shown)


nathanchance:
Understood, thanks for the clarification and review!

PR #79, opened 2020-04-05
README: Update a few package lists

nathanchance:
In my Docker image tests, these couple of packages are necessary for
the build to complete properly.

xz is required after we switched to xz compression on Linux/binutils
tarballs.

texinfo has always been required but I must have left it off by accident
for Fedora since it is there for Debian/Ubuntu. It is a part of the
base-devel group on Arch Linux so no need to update it.


nathanchance:
Thanks for the review as always!

PR #80, opened 2020-04-14
Revert "build-llvm.py: Only enable assertions on the final build"

nathanchance:
This reverts commit 04630c0621e607b260cbc12073feeaf2d13c96c1.

This was actually a bad idea. If assertions are not enabled when doing
PGO, they cannot be taken into account when optimizing hot paths.

Sample kernel build times with a PGO'd clang without assertions:

```
arm32 allyesconfig: 10m 48s
arm64 allyesconfig: 12m 21s
x86_64 allyesconfig: 14m 30s
```

Sample kernel build times with a PGO'd clang with assertions without
this patch:

```
arm32 allyesconfig: 14m 57s
arm64 allyesconfig: 17m 51s
x86_64 allyesconfig: 19m 24s
```

Sample kernel build times with a PGO'd clang with assertions with this
patch:

```
arm32 allyesconfig: 12m 24s
arm64 allyesconfig: 13m 55s
x86_64 allyesconfig: 15m 53s
```

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -703,6 +703,11 @@ def stage_specific_cmake_defines(args, dirs, stage):
         if args.build_type == "Release":
             defines['LLVM_ENABLE_WARNINGS'] = 'OFF'
 
+        # Build with assertions enabled if requested (will slow down compilation
+        # so it is not on by default)
+        if args.assertions:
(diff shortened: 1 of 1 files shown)


nathanchance:
Thanks for the review!

Issue #93, opened 2020-05-22
Allow empty string in --clang-vendor option


dileks:
Currently it is not possible to pass an empty string to `--clang-vendor` option.

This brutal method works:
```
[ build-llvm.py ]

def parse_parameters(root_folder):
...
-                        default="ClangBuiltLinux")
+                        default="")
```

I have opened a separate issue (see my report in [1]).

[1] https://github.com/ClangBuiltLinux/tc-build/issues/92#issuecomment-632489729

nathanchance:
```
$ ./build-llvm.py --clang-vendor ""
```

works fine for me...

```
$ rg CLANG_VENDOR build/llvm/stage1
build/llvm/stage1/CMakeCache.txt
228:CLANG_VENDOR:STRING=
231:CLANG_VENDOR_UTI:STRING=org.llvm.clang
```

dileks:
Hmm, I did a rebuild with ccache-cache. I will try w/o ccache-cache and report.

dileks:
Re-checked w/o ccache-cache and 2nd `build-llvm.py` line (see below):
```
$ grep -nr CLANG_VENDOR build/stage1/CMakeCache.txt
232:CLANG_VENDOR:STRING=
235:CLANG_VENDOR_UTI:STRING=org.llvm.clang
```

I added to my own build-script:
```
# Set clang-vendor string (default: ClangBuiltLinux)
##CLANG_VENDOR="dileks"
CLANG_VENDOR="''"

BASIC_OPTS="--build-type ${BUILD_TYPE} -p ${PROJECTS} -t ${ARCH_TARGETS}"
BASIC_OPTS="${BASIC_OPTS} --clang-vendor ${CLANG_VENDOR}"
```
Before I had `CLANG_VENDOR=""`. That did not work.

dileks:
Maybe it is worth to add a comment that an empty string is allowed.
```
  --clang-vendor CLANG_VENDOR
                        Add this value to the clang version string (like "Apple clang version..." or
                        "Android clang version..."). Useful when reverting or applying patches on top
                        of upstream clang to differentiate a toolchain built with this script from
                        upstream clang or to distinguish a toolchain built with this script from the
                        system's clang. Defaults to ClangBuiltLinux.
```
While looking at the above help text, it might be shorter to write:
```
Useful when reverting or applying patches on top of upstream clang to distinguish/differentiate a toolchain built with this script from upstream or system's clang.
```

[...]

PR #94, opened 2020-05-26
build-llvm.py: Clarify that '--clang-vendor' supports an empty string

nathanchance:
Which is ultimately just the default value in cmake. We chose to change
it because we do a decent amount of configuration to it but there is
nothing fundamentally wrong with removing it.

Closes: #93


dileks:
Looks good to me.
Thanks @nathanchance

PR #111, opened 2020-09-01
Some small documentation fixes

nathanchance:
Each commit should be fairly self explanation.


msfjarvis (approved):
LGTM

dileks:
Both patches look good to me.

nathanchance:
Right as I was about to do it :) thanks for the review both of you!

Issue #132, opened 2021-01-19
[request]Specify the number of compilation threads


zixijian:
When compiling in the chroot container of the mobile phone, the heat is severe and the memory is insufficient.
It is recommended to specify the number of threads. A reasonable number of threads will neither cause fever or frequency reduction nor memory explosion.

nathanchance:
If I understand your message correctly (apologies, I am using Google Translate), can you please give this a shot and see if it works?

https://github.com/nathanchance/tc-build/commit/d067c30d553b7d544c4b526339681fb6d21b61ce

To be fair, I would not personally recommend compiling LLVM on your phone :)

zixijian:
> If I understand your message correctly (apologies, I am using Google Translate), can you please give this a shot and see if it works?
> 
> [nathanchance@d067c30](https://github.com/nathanchance/tc-build/commit/d067c30d553b7d544c4b526339681fb6d21b61ce)
> 
> To be fair, I would not personally recommend compiling LLVM on your phone :)

thanks，it worked.

Frankly speaking, mobile phone compilation is not a big problem. 
I have completed the compilation without specifying threads, and compiled the Android kernel using the tool chain.
That's Why I gave feedback and suggestions on my situation.

PR #133, opened 2021-01-19
build-llvm.py: Add defines options

nathanchance:
Give users the flexibility to define the values that they want so that
it is not necessary to continue to pollute the script with various
flags.

For example, if a user wants to limit the number of compile and link
jobs that ninja can do at one time:

```
$ ./build-llvm.py -D LLVM_PARALLEL_COMPILE_JOBS=2 LLVM_PARALLEL_LINK_JOBS=2
```

Closes: https://github.com/ClangBuiltLinux/tc-build/issues/132

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -143,6 +143,18 @@ def parse_parameters(root_folder):
                         """),
                         type=str,
                         default="ClangBuiltLinux")
+    parser.add_argument("-D",
+                        "--defines",
+                        help=textwrap.dedent("""\
+                        Specify additional cmake values. These will be applied to all cmake invocations.
+
+                        Example: -D LLVM_PARALLEL_COMPILE_JOBS=2 LLVM_PARALLEL_LINK_JOBS=2
+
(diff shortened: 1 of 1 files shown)


ojeda:
The number of link jobs is particularly concerning when building LLVM due to memory usage -- having a separate option for that (rather than using `jobs` for both) would be useful.

MaskRay:
Agree that LLVM_PARALLEL_LINK_JOBS probably should be separate. On a small machine more than 2 link jobs can use up the ram. I think having a way to specify literal -D cmake values will be useful, e.g. just `-D LLVM_PARALLEL_LINK_JOBS=2`.

nathanchance:
I implemented Fangrui's suggestion as this is not the first time I have thought that a `-D` option would be useful!

ojeda (approved):
Looks good!

Good idea including the link to the values, I'm sure a lot of users will appreciate it.

nathanchance:
Thanks for the review!

Issue #123, opened 2020-09-24
pull-request: Document yapf requirements to format Python code


dileks:
When dealing with issue #118 I wanted to offer a pull-request.

Unfortunately, the pull request showed errors in the `yapf` checks - clicking on the `details` has shown no hints.
Desired is the `details` should show the errors.

This was introduced by commit 41cc3c353f40 ("tc-build: Add GitHub Actions workflow").

Here on my Debian system I needed...
```
root# apt-get update 
root# apt-get install python3-yapf yapf3
```

...and checked the format of the `utils.py` file:
```
$ cd tc-build.dileks-github
$ yapf3 -i -vv utils.py
Reformatting utils.py
```

Unsure, if there is more to be installed locally to fullfill any other checks before sending a pull-request.

It is up to you to document whereever you like.
Personally, I prefer `README.md`.

If there are other smart solutions like activating the required GitHub workflows in the own private GitHub account, please let me know and/or document it, too.

[1] https://github.com/ClangBuiltLinux/tc-build/issues/118
[2] https://github.com/ClangBuiltLinux/tc-build/commit/41cc3c353f40f15210587389e0fe611bbb3bf7ea

dileks:
The version I used:
```
$ yapf3 --version
yapf 0.29.0
```

dileks:
A`documentation` label for the `tc-build` `issue-tracker` would be helpful and useful.
It looks like I have not the permissions to create a `new label`.
If you think it is a good idea, feel free to do so.

PR #135, opened 2021-01-20
README: Add a contributing section

nathanchance:
There are some presubmit checks that we do that make it a little hard
for some people to do drive by submissions without guidance. Add a small
section to clarify those and what is expected.

Closes: https://github.com/ClangBuiltLinux/tc-build/issues/123


msfjarvis (approved):
Might be worthwhile to add a `Makefile` to run the checks instead so these can just be pared down to "Install yapf and shfmt, then run `make check`".

nathanchance:
That is a fair point, I'll add it to my backlog!

PR #147, opened 2021-05-06
kernel/build.sh: Update name of source flag and variable

nathanchance:
If a user has SRC_FOLDER defined and exported in their environment, the
script will not work properly. Choose a slighly more specific name to
avoid this and adjust the flag to match.


msfjarvis (approved):

nathanchance:
Thanks for the review as always!

PR #162, opened 2021-06-16
kernel/build.sh: Fix patch application with two PGO kernel runs

nathanchance:
If a user runs `build-llvm.py` with `--pgo kernel-{def,allmod}config`, the
script will wait for their input because a patch has already been
applied:

```
patching file arch/powerpc/include/asm/barrier.h
Reversed (or previously applied) patch detected!  Assume -R? [n]
```

The script should be totally non-interactive. Unfortunately, GNU patch
returns 1 when `patch -N` detects a previously applied patch so we need
to wrap its output and parse it for this phrase, which is not fatal in
our context. Every other message is treated as fatal and printed in
full.

`git apply` is another option but it does not appear to work when patching
untracked files, even with `--no-index` and `--unsafe-paths`.

Fixes: 67f40f7 ("kernel/build.sh: Update patching logic")

### DIFF
--- a/kernel/build.sh
+++ b/kernel/build.sh
@@ -79,6 +79,20 @@ function setup_up_path() {
     [[ -n ${PATH_OVERRIDE} ]] && export PATH=${PATH_OVERRIDE}:${PATH}
 }
 
+# Turns 'patch -N' from a fatal error to an informational message
+function apply_patch {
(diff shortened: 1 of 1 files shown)


msfjarvis (approved):

nathanchance:
Thanks for the review!

PR #175, opened 2021-11-01
kernel: Update to Linux 5.15

dakkshesh07:


msfjarvis (approved):
LGTM, verified the checksum independently.

nathanchance:
Did you test this with `--pgo`? I don't believe this is going to build because of the introduction of `CONFIG_WERROR`.

dakkshesh07:
> Did you test this with `--pgo`? I don't believe this is going to build because of the introduction of `CONFIG_WERROR`.

Yup, just tried the build with pgo and the build failed with werror errors

nathanchance:
We have a couple of different options to avoid this:

1. `KCFLAGS=-Wno-error` in `kernel/build.sh`, which I believe will work.
2. We patch the kernel to remove all instances of `-Werror`.
3. We add a patch series that cleans up all the warnings.

My vote would be 1 or 2, as those are simpler. If you want to take a crack at it, go ahead. Otherwise, I will work on this later today.

torvic9:
ThinLTO + PGO + 5.15 compiles correctly here on main branch [1], with dylib linking.
I didn't run the `--check-targets` though.

[1] `clang version 14.0.0 (https://github.com/llvm/llvm-project 82e0eb22af13df4c97026fcc10387b9370f180c3)`

nathanchance:
I think the `-Werror` breakage will only show up with the `allmodconfig` PGO training (and maybe `x86_64_defconfig` because of the warning you reported in `arch/x86/kvm`, which enables `-Werror` for its subdirectories).

torvic9:
Yes, I think you need at least:

```
KVM: x86: avoid warning with -Wbitwise-instead-of-logical
drm/i915: Avoid bitwise vs logical OR warning in snb_wm_latency_quirk() 
platform/x86: thinkpad_acpi: Fix bitwise vs. logical warning
lib: zstd: Add cast to silence clang's -Wbitwise-instead-of-logical
```

That's what I've come across so far.

PR #176, opened 2021-11-02
Increase number of vp counters per site with PGO

torvic9:
When doing a PGO kernel-defconfig build, the following warning pops up:

`Unable to track new values: Running out of static counters. Consider using option -mllvm -vp-counters-per-site=<n> to allocate more value profile counters at compile time.`

Use the cmake define `LLVM_VP_COUNTERS_PER_SITE` to increase counters to 6 when running the instrumentation stage.
This is an arbitrary value, a lower number may be sufficient.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -956,6 +956,7 @@ def stage_specific_cmake_defines(args, dirs, stage):
         if instrumented_stage(args, stage):
             defines['LLVM_BUILD_INSTRUMENTED'] = 'IR'
             defines['LLVM_BUILD_RUNTIME'] = 'OFF'
+            defines['LLVM_VP_COUNTERS_PER_SITE'] = '6'
 
         # If we are at the final stage, use PGO/Thin LTO if requested
         if stage == get_final_stage(args):


nathanchance:
Will this actually do anything? From [`llvm/cmake/modules/HandleLLVMOptions.cmake`](https://github.com/llvm/llvm-project/blob/d1fdd745d510f40d8741d44ce39f5ae24ee7f91a/llvm/cmake/modules/HandleLLVMOptions.cmake#L984-L992):

```
    # Set this to avoid running out of the value profile node section
    # under clang in dynamic linking mode.
    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND
        CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 11 AND
        LLVM_LINK_LLVM_DYLIB)
      append("-Xclang -mllvm -Xclang -vp-counters-per-site=${LLVM_VP_COUNTERS_PER_SITE}"
        CMAKE_CXX_FLAGS
        CMAKE_C_FLAGS)
    endif()
```

I interpret that as `LLVM_VP_COUNTERS_PER_SITE` is only taken into account when `LLVM_LINK_LLVM_DYLIB` is set, which we do not currently do (and I don't think we want to either, as there are [performance implications](https://lore.kernel.org/r/CAKwvOdk0nxxUATg2jEKgx4HutXCMXcW92SX3DT+uCTgqBwQHBg@mail.gmail.com/)).

I guess that these flags could be added to `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` but I find it odd that you are seeing these warnings as I build with `--pgo kernel-{def,allmod}config` without seeing them. Perhaps something else is going on?

torvic9:
I do build with `LLVM_LINK_LLVM_DYLIB` because the build is faster and size is smaller.

I thought (some of) the performance implications were solved with https://reviews.llvm.org/D102090 and https://reviews.llvm.org/D102453 ?

nathanchance (approved):
I am fine with merging this as it does not impact anyone not using `LLVM_LINK_LLVM_DYLIB`, which the script does not do by default.

PR #178, opened 2021-12-14
Improvements around non-x86_64 native use

nathanchance:
This pull request removes all assumptions that these scripts are being run on x86_64 hardware. It has been verified on both AArch64 and x86_64 hardware. See the individual commits for the improvements.


msfjarvis (approved):

nathanchance:
Thanks for the review!

PR #187, opened 2022-04-21
tc-build: Initial BOLT support

nathanchance:
From bolt/README.md:

"BOLT is a post-link optimizer developed to speed up large applications.
It achieves the improvements by optimizing application's code layout
based on execution profile gathered by sampling profiler, such as Linux
perf tool."

Wire up support for BOLT into tc-build by generating a profile through
perf if it is support or BOLT's instrumentation if not. The profile is
generated by building a kernel with the final clang binary. The kernel
will either be the host architecture's defconfig if it is supported in
the lists of targets or the first target's defconfig.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -5,6 +5,7 @@ import argparse
 import datetime
 import glob
 import pathlib
+import platform
 import os
 import subprocess
 import shutil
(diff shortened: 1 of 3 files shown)


ptr1337:
@nathanchance 

I did used for my bash script the following check, which is quite simply and easy:

```perf record -e cycles:u -j any,u -- sleep 1 &>/dev/null;
if [[ $? == "0" ]]; then
  echo "BOLTING with Profile!"
  ./build_stage3-bolt.bash || (echo "Optimizing Stage2-Toolchain further with llvm-bolt failed!"; exit 1)
else
  echo "Optimizing Stage2-Toolchain without a perf record profile"
  ./build_stage3-bolt-without-profile.bash || (echo "Optimizing Stage2-Toolchain with bolt failed!"; exit 1)
fi```

Also in my eyes instrumenting clang with building the kernel would blow up the needed space for the profile completely. I did personally just built clang like described here:
https://github.com/llvm/llvm-project/blob/main/bolt/docs/OptimizingClang.md#optimizing-clang-with-bolt
and the perf.data already got around 40 GB. If profiling clang with building the kernel it would be a massive profile which probably cant even me aggregated.

Also if perf was used to create the optimized binary, i wouldnt remove the PGO binary, so just rename them into clang.org;clang.pgo for example. So its possible to compare them after compiling. 

Regards

**Comment by nathanchance** [comment]
> I did used for my bash script the following check, which is quite simply and easy:

Thanks, that is basically what I did for the `perf` support that I wrote yesterday.

> Also in my eyes instrumenting clang with building the kernel would blow up the needed space for the profile completely. I did personally just built clang like described here: https://github.com/llvm/llvm-project/blob/main/bolt/docs/OptimizingClang.md#optimizing-clang-with-bolt and the perf.data already got around 40 GB. If profiling clang with building the kernel it would be a massive profile which probably cant even me aggregated.

Hmm, yes, thank you for mentioning that. I saw a similar issue with building five kernels (`arm64 defconfig`, `arm multi_v5_defconfig`, `arm aspeed_g5_defconfig`, `arm multi_v7_defconfig`, and `x86_64 defconfig`): `perf.data` was around 80GB or so and that entire file was loaded into memory by `perf2bolt` calling `perf script`, which took so long that I cancelled the build. Instead, we'll just build the host architecture's defconfig if it is in the list of targets or the first target's defconfig, which should still yield a decent profile. I'll benchmark it before pushing though.

> Also if perf was used to create the optimized binary, i wouldnt remove the PGO binary, so just rename them into clang.org;clang.pgo for example. So its possible to compare them after compiling.

I'll consider this. I would rather not keep around two full `clang` binaries, as that is quite a bit of wasted space, but maybe I'll just move the original PGO'd `clang` into the build folder, which is where all of the other artifacts reside.

**Comment by nathanchance** [comment]
Alright, I think this is ready for formal review...

**Review by nickdesaulniers — APPROVED** [review]

**Comment by nathanchance** [comment]
Thanks for the review! If there are any other comments after this is merged, I'll address them in a follow-up PR.

**Comment by dileks** [comment]
LLVM `release/14.x` Git branch includes and version `14.0.4` will have this:
```
commit 7a42b2fd5be3f4141bd0db3c5efa2ebb74ef658d
"[BOLT] Compact legacy profiles"
```
[1] https://github.com/llvm/llvm-project/commit/7a42b2fd5be3f4141bd0db3c5efa2ebb74ef658d

PR #191, opened 2022-06-03
Update to Linux 5.18 and update known good revision

nathanchance:
Qualified with https://github.com/nathanchance/llvm-kernel-testing on
an aarch64 and x86_64 system.


nickdesaulniers (approved):
sha's match on my end

PR #203, opened 2022-09-14
tc-build: Flush print commands to ensure output is properly ordered

nathanchance:
Otherwise, when running non-interactively (such as through GitHub
Actions), the output from these print statements might not appear before
the output of subprocess.run(), making it hard to decipher what has gone
wrong.


stephenhines (approved):

nathanchance:
Thanks for the review!

PR #209, opened 2022-11-18
tc-build: Eliminate os.path usage and more f-strings

nathanchance:
This modernizes build-llvm.py a little bit more.


msfjarvis (approved):

nathanchance:
I found a bug when testing with PGO. I'll have a fix shortly, just waiting on my builds to finish. Don't merge until I have pushed that fix please.

PR #211, opened 2022-12-15
build-llvm.py: Avoid LLVMgold unless using ld.gold and lto

kees:
Don't build LLVMgold if we don't absolutely have to.

Signed-off-by: Kees Cook <keescook@chromium.org>

PR #212, opened 2022-12-16
tc-build: Stop building the gold plugin

nathanchance:
The only reason that we built the gold plugin in the first place was to
support LTO on the Pixel 3, which used ld.gold instead of ld.lld, unlike
Android 4.14+ and mainline's LTO implementation.

The Pixel 3 has been EOL for several months now, so rip out all the gold
plugin enablement. With that, all the binutils functions in utils can be
moved into build-binutils.py directly, as build-llvm.py does not need to
know about binutils anymore.

Cc @kees, in lieu of #211.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -620,10 +620,9 @@ def ref_exists(repo, ref):
     return True
 
 
-def fetch_llvm_binutils(root_folder, llvm_folder, update, shallow, ref):
+def fetch_llvm(llvm_folder, update, shallow, ref):
     """
-    Download llvm and binutils or update them if they exist
-    :param root_folder: Working directory
+    Download llvm if it does not exist, update it if it does
     :param llvm_folder: llvm-project repo directory
     :param update: Boolean indicating whether sources need to be updated or not
     :param ref: The ref to checkout the monorepo to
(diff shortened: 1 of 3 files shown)


build-llvm.py
```diff
@@ -1073,11 +1066,6 @@ def stage_specific_cmake_defines(args, dirs, stage):
             if key not in str(args.defines):
                 defines[key] = ''
 
-        # For LLVMgold.so, which is used for LTO with ld.gold
-        defines['LLVM_BINUTILS_INCDIR'] = dirs.root_folder.joinpath(
-            utils.current_binutils(), "include")
-        defines['LLVM_ENABLE_PLUGINS'] = 'ON'
```
stephenhines, line 1079: I suspect that this is still going to default to true/ON for your builds.

stephenhines (approved):
This would be faster to review if the refactoring were separated from the part that just turns off building the plugin.

nathanchance:
> This would be faster to review if the refactoring were separated from the part that just turns off building the plugin.

Good point, sorry about that. I'll keep that in mind for the future.

> I suspect that this is still going to default to true/ON for your builds.

Indeed, we'll inherit the default of `LLVM_ENABLE_PIC`, which is `ON`. We could look into turning it off now that we don't need it but I can do that in a follow up PR. Thanks for the review!

kees:
Thank you! This works for me. :)

PR #216, opened 2022-12-16
build-llvm.py: Add option to reduce CMake output

kees:
CMake dominates the number of lines spewed during a build. Quiet the output when running with --quiet-cmake.

Signed-off-by: Kees Cook <keescook@chromium.org>


nathanchance (approved):
I like it!

PR #218, opened 2023-01-14
utils.py: Update to binutils 2.40

najahiiii:
Signed-off-by: Ahmad Thoriq Najahi <najahi@zephyrus.id>

### DIFF
--- a/build-binutils.py
+++ b/build-binutils.py
@@ -18,7 +18,7 @@ def current_binutils():
     Simple getter for current stable binutils release
     :return: The current stable release of binutils
     """
-    return "binutils-2.39"
+    return "binutils-2.40"
 
 
 def download_binutils(folder):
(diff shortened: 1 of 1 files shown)


nathanchance (approved):
Thanks for the patch!

nathanchance (changes requested):
Actually, it looks like `yapf` is not happy with the URL change. Please run it locally or manually apply the diff that CI suggests.

najahiiii:
Updated

nathanchance:
Let's use an f-string for this to satisfy both `flake8` and `yapf` (hopefully...):

```
f"https://sourceware.org/pub/binutils/releases/{binutils_tarball.name}"
```

najahiiii:
Updated, passed both pylint check locally.

nathanchance (approved):

PR #219, opened 2023-01-20
Add support for LoongArch

nathanchance:
This pull request adds support for targeting LoongArch, as it may soon be possible to build the LoongArch Linux kernel with clang:

https://github.com/ClangBuiltLinux/linux/issues/1787

The first two patches just clean up a couple things I noticed while writing this series.

The third patch adds support for validating the targets that the user passes; this is largely tangential to this series but `get_all_targets()` is needed for checking if LoongArch can be enabled in LLVM and it is a logical extension to validate the user's input as well.

Finally, the last three patches wire up LoongArch into `build-binutils.py`, `build-llvm.py`, and `kernel/build.sh` (as a no-op for now).

If there are any questions or comments, please let me know.

### DIFF
--- a/build-binutils.py
+++ b/build-binutils.py
@@ -91,10 +91,11 @@ if args.build_folder:
 else:
     build_folder = Path(tc_build_folder, 'build/binutils')
 
+default_targets = bsm.default_targets()
 if args.targets:
-    targets = bsm.default_targets() if 'all' in args.targets else set(args.targets)
+    targets = default_targets if 'all' in args.targets else set(args.targets)
 else:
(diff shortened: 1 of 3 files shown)


nathanchance:
I am converting this to a draft until #226 is merged, as this PR was the catalyst for that and I think this can be done much cleaner after that change.

nathanchance:
This should be functionally equivalent to the previous iteration but it is rebased on the rewrite that occurred in #226.

nickdesaulniers (approved):

msfjarvis (approved):

Issue #221, opened 2023-01-22
Display total build-time


dileks:
After I have built a stage1-only `LLVM-15.0.7` toolchain, I did a ThinLTO + PGO optimized (for `Linux v6.2-rc4` x86_64-defconfig).

Is it possible to create a `timestamp before and after the build` to have a `total build-time`?

Here the durations for the single stages, kernel and tests:
```
$ egrep '== Building|LLVM build duration:|Testing Time|real' archives/dileks-llvm-toolchain-15.0.7-5-g2c432a3d6ea8-thinlto_pgo_optimized-KCFI/log_tc-build.txt 
522:== Building LLVM stage 1 ==
4425:LLVM build duration: 0:12:41 <--- Rebuild from ccache (stage1-only: LLVM build duration: 4:09:08)
4742:== Building LLVM stage 2 ==
8321:LLVM build duration: 3:11:17
8324:== Building PGO profiles ==
8339:== Building kernels (6.2.0-rc4) ==
8357:real       66m53.470s
8676:== Building LLVM stage 3 ==
12445:Testing Time: 139.26s
12473:Testing Time: 828.88s
12479:LLVM build duration: 11:58:59
```
Counting together should be approx. 16h30m.

Total time matches:
```
$ cat start-build.txt dileks     17856   17853  0 21:52 pts/1    00:00:00 python3 ./build-llvm.py --no-update --build-type Release -p clang;lld -t X86;BPF --clang-vendor dileks -B /home/dileks/src/llvm/build -I /opt/llvm --check-targets clang lld --lto thin --pgo kernel-defconfig -L /home/dileks/src/linux/git -D LLVM_PARALLEL_LINK_JOBS=1 --show-build-commands

$ ls -alth /opt/llvm/ | head -3
insgesamt 36K
drwxr-xr-x 2 dileks dileks 4,0K 21. Jan 14:27 bin
-rw-r--r-- 1 dileks dileks    2 21. Jan 14:27 gitignore
```

Thanks.

nathanchance:
Sure, try out https://github.com/ClangBuiltLinux/tc-build/pull/222.

dileks:
Thanks for the quick patch!

I have a marathon-session of approx. 24 hours with building LLVM toolchain + Linux-6.2-rc5, so today not really in the mood to do any more builds.
I simply trust you.

Hmm, can I use it to check/see the total build-time of binutils-2.40?
build-binutils.py handles this, don't think so.

[...]

PR #222, opened 2023-01-22
Print total script durations

nathanchance:
Closes: https://github.com/ClangBuiltLinux/tc-build/issues/221


msfjarvis (approved):

nathanchance:
I added a better (IMO, that is) duration formatting function and applied it to `build-binutils.py` as well.

PR #225, opened 2023-01-26
Update kernel to 6.1.7 and bump known good revision

nathanchance:
- kernel: Update to Linux 6.1.7
- kernel: Add patch to workaround __read_overflow() error with s390 allmodconfig
- build-llvm.py: Update known good revision


msfjarvis (approved):
The patch feels a bit icky but I guess ultimately irrelevant to tc-build.

nathanchance:
> The patch feels a bit icky but I guess ultimately irrelevant to tc-build.

I do not disagree :)

Unfortunately, `__read_overflow()` is always an error, even with -Werror, so there is not really a way to work around this from the command line, other than just not building that driver with s390. That could be the better option but it requires a little more plumbing that I am not sure is worth it, given how infrequently we should hopefully run into this situation.

PR #226, opened 2023-01-30
tc-build: Rewrite

nathanchance:
This is currently a work in progress: I believe `build-llvm.py` is feature complete but I still need to port `build-binutils.py` over and rewire the basic CI. As `build-llvm.py` is the main point of this repository, I wanted to push this for initial testing and review as soon as possible.

There are quite a few breaking changes that I made in the pursuit of simplicity and maintainability. See the commit message of the rewrite commit for the full list, I do not want to be constantly editing the commit message and the message of this pull request as things pop up that need to be mentioned. The rewrite commit hash is currently 2fb5c1b9a4b6940ae02291341499543df3ad941b but it may change throughout review.

See that same commit message for some of the reasoning behind this; I think this is much more maintainable as everything is clearly separated and compartmentalized now. I apologize that the rewrite is a single commit instead of a series, I hope it will not be a burden to review. I tried to keep everything easily separated so that each area was not overwhelming to understand.

I welcome any and all bug reports but please keep them here rather than in the issues tab until this is merged.

cc @ConchuOD @dileks @kees, as I know you use the script at times, so I would like to avoid breaking your use cases.


ConchuOD:
I gave my regular old toolchain build a go and that's working again. I haven't tried the PGO one yet, I'll give it a go tomorrow (if I remember!)

ConchuOD:
I'm not overly sold on the deprecation & reuse of arguments, but I don't really view the backwards compat of this script as being particular important. -b -> -r, -B -> -b might do something weird for some people I suppose. I just can't bring myself to care?

nathanchance:
> I gave my regular old toolchain build a go and that's working again. I haven't tried the PGO one yet, I'll give it a go tomorrow (if I remember!)

Appreciate it!

> I'm not overly sold on the deprecation & reuse of arguments, but I don't really view the backwards compat of this script as being particular important. -b -> -r, -B -> -b might do something weird for some people I suppose. I just can't bring myself to care?

Yeah, I am not the biggest fan of it either but I think the amount of options that the script has accrued over time is a little on the larger side so if I am already rewriting the scripts, it seems reasonable to do a one-time deprecation period.

Perhaps we could add a one time notice to the script that will notify the user of the new options like:

```
if not (flag_file := Path(tc_build_folder, '.user_notified_of_options_break')).exists():
    tc_build.utils.print_warning('The options to this script has changed! Please review them with "--help"...')
    time.sleep(10)
    file_file.touch()
```

> I would like to see that in an own Git branch.

I am not sure I understand this comment. Are you saying that even after this is merged, it should live a different branch than main?

nathanchance:
Alright, I believe this is ready for full testing, as `build-binutils.py` has been ported over.

As I mentioned above, if there are concerns with backwards compatibility, we could warn the user once then write a git ignored file to disk to avoid warning in the future. At the same time, if the user is pulling a new update, I would expect them to glance at the commit log to see what new changes they are getting. I do not feel strongly about it, I'll do whatever people feel is right.

dileks:
So you pushed several time your pull-226-patchset.

I would like to see:

tcbuild.github#Rewrite-v1
tcbuild.github#Rewrite-v2

...as branches beside main.

Easier for me to compare with main v1 v2 etc.
But might be my gusto - preferred style of working with Git.

`git pull origin pull/226/head` will take the latest one - and the previous versions?

BTW, `git describe`?
Might worth to give your rewrite a v1.0 :-)?

nathanchance:
> So you pushed several time your pull-226-patchset.
> 
> I would like to see:
> 
> tcbuild.github#Rewrite-v1 tcbuild.github#Rewrite-v2
> 
> ...as branches beside main.
> 
> Easier for me to compare with main v1 v2 etc. But might be my gusto - preferred style of working with Git.

GitHub's UI has `Compare` buttons next to my force pushes if you want to see the diffs between my pushes but the rewrite has not been completed until this most recent push. The first force push was for the `build-binutils.py` rewrite and the second force push was updating `ci.sh`.

> `git pull origin pull/226/head` will take the latest one - and the previous versions?

The previous versions were either buggy or incomplete, as I mention above. Consider this most recent push as `v1.0`, I have marked it as ready for review.

Any future changes that are needed for this rewrite will be done as individual commits on top.

ConchuOD:
Something I just noticed today Nathan, llvm and bintuils don't use the same args for providing a build dir - is this an opportunity to align the two?
llvm uses -b & binutils uses -B (as -b is for the binutils source)

nathanchance:
> is this an opportunity to align the two?

Sure thing, might as well when making breaking changes: a7edf39f4d1a6629cbec5ab8054c86134f3d3e11

nathanchance:
I have pushed away the `ruff` changes for the moment, nothing has functionally changed with the pull request.

tc_build/kernel.py
```diff
@@ -0,0 +1,401 @@
...
+        # This is a little convoluted :/
+        # The overall idea here is to avoid duplicating builds, so the
+        # matrix consists of a series of configuration targets ("defconfig",
+        # "allmodconfig", etc) and a list of LLVM targets to build for each
+        # configuration target. From there, this block filters out the
+        # architectures that cannot build their "all*configs" with clang, so
+        # they are duplicated if both "defconfig" and "allmodconfig" are
+        # requested.
+        for config_target, llvm_targets in self.matrix.items():
+            if config_target == 'defconfig':
```
nickdesaulniers, line 312: To avoid more and more indentation, when there's no else clause you could always invert the conditional and continue early.

```python
if config_target != 'defconfig':
    continue
for llvm_target in llvm_targets:
```
that way you code doesn't keep creeping to the right.

tc_build/llvm.py
```diff
@@ -0,0 +1,545 @@
...
+            's390x': 'SystemZ',
+            'x86_64': 'X86',
+        }
+        return uname_to_llvm.get(platform.machine())
+
+    def host_target_is_enabled(self):
+        if 'all' in self.targets:
+            return True
+
+        return self.host_target() in self.targets
```
nickdesaulniers, line 299:
```python
return 'all' in self.targets or self.host_target() in self.targets`
```

nickdesaulniers: You do this pattern below in `project_is_enabled`.

tc_build/llvm.py
```diff
@@ -0,0 +1,545 @@
...
+        if self.repo.exists():
+            return
+
+        utils.print_header('Downloading LLVM')
+
+        git_clone = ['git', 'clone']
+        if shallow:
+            git_clone.append('--depth=1')
+            if ref != 'main':
+                git_clone.append('--no-single-branch')
```
nickdesaulniers, line 504: Do you mean for this to be conditional on `shallow`?

tc_build/source.py
```diff
@@ -0,0 +1,88 @@
...
+
+            if 'sha256' in self.remote_checksum_name:
+                file_hash = hashlib.sha256()
+            elif 'sha512' in self.remote_checksum_name:
+                file_hash = hashlib.sha512()
+            else:
+                raise RuntimeError(
+                    f"No supported hashlib for {self.remote_checksum_name}, add support for it?")
+            with self.local_location.open('rb') as file:
+                while (data := file.read(131072)):
```
nickdesaulniers, line 53: That's a magical constant number.  Why don't you give it a meaningful identifier?

nickdesaulniers (approved):

tc_build/source.py
nathanchance: Sure thing!

01b65b97a480b3eeeab0c6759d2bfef6836f7189

tc_build/llvm.py
nathanchance: This is in an `if shallow` block already.

tc_build/llvm.py
nathanchance: Ha, good catch! If there is one thing I am good at, it is inconsistency :P

4fd6541ae36f1fdb420c988810dc8b26255c01c2

tc_build/kernel.py
nathanchance: I do not think that suggestion would work for what this block is trying to do but I tried to simplify it in 0c1452ee9c5ea3d21ecfb5c08b682b75a9610f9f to make it clearer, which reduces the indentation.

tc_build/builder.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+
+import shlex
+import shutil
+import subprocess
+
+
+class Folders:
```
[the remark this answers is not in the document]
nathanchance: > It's not super common to have class names be plural.

Understandable.

> Having a class Folder, then having a list of instances of Folder objects with the member or variable identifier of folders would perhaps be more ergonomic.

The whole point of the class is to keep the folders grouped together and easily identifiable. At that point, I might as well just turn this into a dictionary, right?

https://gist.github.com/nathanchance/06016b8ac2afcdd50e2fea6432136b33

I do not really have a strong opinion though.

nathanchance:
I plan to merge this in its current state in a few hours, I think it has had enough time for review and testing. I am happy to address any follow up comments with new pull requests, should there be any. Please file issues for any bugs found after the merge for easy tracking.

MajorP93:
@nathanchance Will there be differences for users? Will the command line arguments be the same?

nathanchance:
@MajorP93 yes, there are some breaking changes, most of which you can read about in 8bc3594fde9585df0cedeeb9e79a64412ad4c686.

You can see the command line arguments for each here to prepare for any changes you may have to make:

https://github.com/ClangBuiltLinux/tc-build/blob/0c1452ee9c5ea3d21ecfb5c08b682b75a9610f9f/build-binutils.py#L14-L67

https://github.com/ClangBuiltLinux/tc-build/blob/0c1452ee9c5ea3d21ecfb5c08b682b75a9610f9f/build-llvm.py#L26-L399

You can see the changes that I had to make for my own personal wrappers here:

https://github.com/nathanchance/env/commit/a8ec5ce5e6b4d3eb8b5ec46247618a8161e78648

MajorP93:
@nathanchance Ok I see. Thanks for the explanation.

PR #228, opened 2023-02-20
build-llvm.py: Fix setting LD fails due to inverted test-check

iRebbok:
When I set `LD` to any legit linker, for example `lld` or `gold`, it prints that it won't work but an incorrect one will go and fail during compilation.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -451,7 +451,7 @@ def linker_test(cc, ld):
     Test to see if the supplied ld value will work with cc -fuse=ld
     :param cc: A working C compiler to compile the test program
     :param ld: A linker to test -fuse=ld against
-    :return: False if the linker supports -fuse=ld, True otherwise
+    :return: 0 if the linker supports -fuse=ld, 1 otherwise
     """
     cc_cmd = [cc, f'-fuse-ld={ld}', '-o', '/dev/null', '-x', 'c', '-']
 
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -566,7 +566,7 @@ def check_cc_ld_variables(root_folder):
         # and we're using gcc, try to use gold
         else:
             ld = "gold"
-            if linker_test(cc, ld):
+            if not linker_test(cc, ld):
                 ld = None
 
     # Print what binaries we are using to compile/link with so the user can
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -462,9 +462,9 @@ def linker_test(cc, ld):
                        input='int main() { return 0; }',
                        text=True)
     except subprocess.CalledProcessError:
-        return False
+        return True
 
-    return True
+    return False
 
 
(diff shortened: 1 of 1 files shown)


msfjarvis (commented):
While we're here please change the comment on line 454 to match the correct expected values instead of the clearly misunderstood 0/1

nathanchance:
That means your compiler invocation is failing. Can you use this to see why?

```diff
diff --git a/build-llvm.py b/build-llvm.py
index ed01ba3..9352f32 100755
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -461,7 +461,11 @@ def linker_test(cc, ld):
                        check=True,
                        input='int main() { return 0; }',
                        text=True)
-    except subprocess.CalledProcessError:
+    except subprocess.CalledProcessError as err:
+        if err.stderr:
+            print(err.stderr)
+        if err.stdout:
+            print(err.stdout)
         return False

     return True
```

nathanchance:
Re-reading the pull request description, I am having a hard time figuring out how the problem initially occurs and how this patch will fix it. You are saying that `linker_test()` returns `False` when a valid `-fuse-ld` value for `LD` is used but it returns `True` when an invalid value is used, so compilation fails later on?

How can `subprocess.run()` not fail in this scenario? That is the point of the diff above. It seems like this patch introduces the very behavior you describe, but I could be misunderstanding something.

iRebbok:
> That means your compiler invocation is failing. Can you use this to see why?
> 
> ```diff
> diff --git a/build-llvm.py b/build-llvm.py
> index ed01ba3..9352f32 100755
> --- a/build-llvm.py
> +++ b/build-llvm.py
> @@ -461,7 +461,11 @@ def linker_test(cc, ld):
>                         check=True,
>                         input='int main() { return 0; }',
>                         text=True)
> -    except subprocess.CalledProcessError:
> +    except subprocess.CalledProcessError as err:
> +        if err.stderr:
> +            print(err.stderr)
> +        if err.stdout:
> +            print(err.stdout)
>          return False
> 
>      return True
> ```

Actually, prints nothing. Either manually executing `/usr/bin/clang -fuse-ld=mold -o /dev/null -x c - <<< 'int main() { return 0; }'` prints nothing and exits with 0 code. Looking at line 457 I clearly see it fails because of inverted return value of `linker_test` tho.

nathanchance:
> because of inverted return value of `linker_test` tho.

Right, I just noticed this, sorry about that :( those should both be `if not linker_test(...)`, care to update the pull request and I will merge it?

nathanchance (approved):
Thanks a lot for the report and fix! If you have a second, could you verify that I avoided this same problem in the rewrite that I have pending in #226? I think I did but it never hurts to have a little more testing :)

iRebbok:
> 

The `rewrite` branch has no similar issue.

PR #229, opened 2023-03-23
build-llvm.py: Switch to dict() over dictionary comprehension

nathanchance:
Something is going wrong with the comprehension, as there is a
ValueError when unpacking the define (which makes little sense):

```
Traceback (most recent call last):
  File ".../tc-build/./build-llvm.py", line 521, in <module>
    defines = {k: v for define in args.defines for1 k, v in define.split('=', 1)}
  File ".../tc-build/./build-llvm.py", line 521, in <dictcomp>
    defines = {k: v for define in args.defines for k, v in define.split('=', 1)}
ValueError: too many values to unpack (expected 2)
```

This can be worked around by just using the `dict()` constructor, which is
simpler to read anyways.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -518,7 +518,7 @@ if args.vendor_string:
     common_cmake_defines['CLANG_VENDOR'] = args.vendor_string
     common_cmake_defines['LLD_VENDOR'] = args.vendor_string
 if args.defines:
-    defines = {k: v for define in args.defines for k, v in define.split('=', 1)}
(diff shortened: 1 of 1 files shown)


msfjarvis (approved):

nathanchance:
> Is `for k, v in define.split('=', 1)` producing more than one value, for some given input?

it might be but it makes literally zero sense because `k, v = define.split('=', 1)` works for the same exact input…

PR #236, opened 2023-06-21
Update README.md

Albert753258:
Without libstdc++-12-dev build fails with
`/usr/bin/ld: cannot find -lstdc++: No such file or directory`

```
CMake Error at /usr/share/cmake-3.22/Modules/CMakeTestCXXCompiler.cmake:62 (message):
  The C++ compiler
    "/usr/lib/llvm-14/bin/clang++"
  is not able to compile a simple test program.
  It fails with the following output:
    Change Dir: /home/albert/rom/tc-build/build/llvm/bootstrap/CMakeFiles/CMakeTmp
    Run Build Command(s):/usr/bin/ninja cmTC_86a61 && [1/2] Building CXX object CMakeFiles/cmTC_86a61.dir/testCXXCompiler.cxx.o
    [2/2] Linking CXX executable cmTC_86a61
    FAILED: cmTC_86a61 
    : && /usr/lib/llvm-14/bin/clang++   CMakeFiles/cmTC_86a61.dir/testCXXCompiler.cxx.o -o cmTC_86a61   && :
    /usr/bin/ld: cannot find -lstdc++: No such file or directory
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    ninja: build stopped: subcommand failed.
  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:47 (project)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,12 @@ These scripts have been tested in a Docker image of the following distributions 
               clang \
               cmake \
               curl \
-              libstdc++-12-dev \
               file \
               flex \
               git \
               libelf-dev \
               libssl-dev \
+              libstdc++-$(apt list libstdc++6 2>/dev/null | grep -Eos '[0-9]+\.[0-9]+\.[0-9]+' | head -1 | cut -d . -f 1)-dev \
               lld \
               make \
               ninja-build \
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ These scripts have been tested in a Docker image of the following distributions 
               clang \
               cmake \
               curl \
+              libstdc++-12-dev \
               file \
               flex \
               git \


nathanchance:
Please keep the list alphabetically sorted.

I don’t think we need the `-dev` package, just `libstdc++6`. Either way, I would like to avoid having a version number in this list because not every version of Ubuntu or Debian has this package. Is there a virtual package that can be installed if the `-dev` version is necessary so that it always works?

Albert753258:
Without dev package build fails. I have `libstdc++6 is already the newest version (12.1.0-2ubuntu1~22.04)`, but build still fails. libstdc++-dev is virtual package
```
sudo apt install libstdc++-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package libstdc++-dev is a virtual package provided by:
  libstdc++-9-dev 9.5.0-1ubuntu1~22.04
  libstdc++-12-dev 12.1.0-2ubuntu1~22.04
  libstdc++-10-dev 10.4.0-4ubuntu1~22.04
  libstdc++-11-dev 11.3.0-1ubuntu1~22.04.1
You should explicitly select one to install.

E: Package 'libstdc++-dev' has no installation candidate
```

nathanchance:
Okay, something like

```
libstdc++-$(apt list libstdc++6 2>/dev/null | grep -Eos '[0-9]+\.[0-9]+\.[0-9]+' | cut -d . -f 1)-dev
```

should work for all Debian and Ubuntu versions. I just tested in Docker images with Debian Bullseye, Buster, and Bookworm and Ubuntu 18.04, 20.04, and 22.04.

Albert753258:
Command you provided working incorrectly on ubuntu 22.04
```
echo libstdc++-$(apt list libstdc++6 2>/dev/null | grep -Eos '[0-9]+\.[0-9]+\.[0-9]+' | cut -d . -f 1)-dev
libstdc++-12 12-dev
```
We should add head -1 to it
```
echo libstdc++-$(apt list libstdc++6 2>/dev/null | grep -Eos '[0-9]+\.[0-9]+\.[0-9]+' | head -1 | cut -d . -f 1)-dev
libstdc++-12-dev
```

nathanchance:
What is the output of `apt list libstdc++6`? There may be a better way to handle this situation.

nathanchance (approved):
Ah multilib, duh :) seems reasonable, thanks a lot!

PR #244, opened 2023-09-23
build-llvm.py: Utilize thin archives

MaskRay:
Copying .o files into archives are wasteful. Utilize thin archives to reclaim wasted space. Use the deprecated -T for compatibility with binutils<2.38 and llvm-ar<14.

See https://maskray.me/blog/2022-01-16-archives-and-start-lib for some notes about thin archives and useless archive symbol tables. In an ideal world where GNU ld did not read archive symbol tables, we could specify -S to save more space.

### DIFF
--- a/tc_build/llvm.py
+++ b/tc_build/llvm.py
@@ -226,6 +226,16 @@ class LLVMBuilder(Builder):
 
         if self.tools.ar:
             self.cmake_defines['CMAKE_AR'] = self.tools.ar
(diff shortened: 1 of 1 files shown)


MaskRay:
Ping:)

nathanchance (approved):
Sorry for the delay, I have been traveling for the last two weeks. This seems to work fine for me locally, I have not tested all different versions of LLVM that the kernel supports but if there are any problems later, I’ll just add a version check or something. Thanks for contributing to tc-build!

PR #246, opened 2023-10-23
tc_build: llvm: Fix distribution_components with disabled projects

nathanchance:
It is possible to use `--projects` to build certain combinations of subprojects, such as just `llvm` or `clang` to build just LLVM or LLVM and clang respectively. When either the `clang` or `lld` subproject is not enabled, there is an error when running `cmake` due to their targets being present in the distributions components list.

Only add them to `distribution_components` if the projects that contain them are enabled.


msfjarvis (approved):
Are the remaining default entries in `distribution_components` guaranteed to be available for all combinations?

nathanchance:
> Are the remaining default entries in `distribution_components` guaranteed to be available for all combinations?

I think `LLVM_BUILD_TOOLS=OFF` would break under the current scheme. I thought about just letting that be but I think it is easy enough to guard against: e916997047aa8197957980adca94c88e77869ec6

msfjarvis (approved):

PR #299, opened 2025-04-13
Add basic support for building Rust

ojeda:
There a lot of options missing and it does not do any fancy kind of build. But it is a start, and it is already useful, e.g. Peter could have used it to test the new KCFI arity flag that requires LLVM 21 but upstream Rust still uses LLVM 20. It could also be useful to potentially/eventually have LLVM+Rust builds in kernel.org that use the exact same LLVM build.

I took the approach that the new script only takes care of building Rust provided an existing LLVM, which seemed simple and clear.

Thus add the basic infrastructure, plus a bit of documentation. Add it to the CI, too.

I successfully built it in a clean Debian 12 and Fedora 41.

A few notes:

  - The script works on its own, but the CI does not pass because `llvm-config --bindir` does not work with the set of flags used in CI for LLVM -- we are missing at least `llvm-config` in the distribution components of the LLVM step, but then also headers like `llvm/Config/llvm-config.h`.

    I am not sure how far you would want to modify the LLVM side or the `do_llvm()` step, so I decided to send this with the failing CI so that you see it.

  - If `LLVM_INSTALL_UTILS` is enabled (to get `FileCheck` installed, which is used by the Rust build system for testing), then we can leave `codegen-tests` to its default (enabled) in the Rust configuration.

    But perhaps we want to leave those tests anyway as the equivalent of the `check-*`s in LLVM, i.e. disabled by default even if the Rust configuration enables them by default, and let the user enable them.

  - Rust uses submodules and so on and manages them on its own if allowed, so I just did that. I didn't add/test shallow clones, so I didn't add them.

  - This does nothing about the LLVM patches that Rust may carry (until they get upstreamed). One could use `--rust-folder` to point to a checkout of the branch that Rust uses, and perhaps we should add some documentation about it, or even provide an option in `./build-llvm.py` to build such a branch, but I didn't try/test anything related to that yet.

  - We probably want to offer eventually `--quiet-x`, `--no-locked-deps`, `--tools`, `--codegen-tests`, `--docs`, etc.

  - After this, we could add a `--build-bindgen` here or perhaps a new `./build-bindgen.py`, so that we can easily get a full toolchain for the kernel. It is really just a one liner once one has the Rust toolchain, so it is not too important.

  - Some refactoring could be done so that both `./build-llvm.py` and `./build-rust.py` reuse certain bits, e.g. `show_install_info()`.

I tried to follow the LLVM script as closely as possible and "copy" the style etc., but there may be mistakes -- please feel free to take this and modify it as you see fit, of course.

I hope this helps!

Cc: Peter Zijlstra (by email)

### DIFF
--- a/build-rust.py
+++ b/build-rust.py
@@ -84,7 +84,7 @@ parser.add_argument('-r',
                     does not own.
 
                     '''),
-                    default='master',
+                    default='main',
                     type=str)
 parser.add_argument('--show-build-commands',
                     help=textwrap.dedent('''\
--- a/build-rust.py
+++ b/build-rust.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+# pylint: disable=invalid-name
+
+from argparse import ArgumentParser, RawTextHelpFormatter
+from pathlib import Path
+import textwrap
+import time
+
+import tc_build.utils
+
+from tc_build.rust import RustBuilder, RustSourceManager
+
+# This is a known good revision of Rust for building the kernel
+GOOD_REVISION = '69b3959afec9b5468d5de15133b199553f6e55d2'
+
+parser = ArgumentParser(formatter_class=RawTextHelpFormatter)
+clone_options = parser.add_mutually_exclusive_group()
+
+parser.add_argument('--debug',
+                    help=textwrap.dedent('''\
+                    Build a debug compiler and standard library. This enables debug assertions,
+                    debug logging, overflow checks and debug info. The debug assertions and overflow
+                    checks can help catch issues when compiling.
+
+                    '''),
+                    action='store_true')
+parser.add_argument('-b',
+                    '--build-folder',
+                    help=textwrap.dedent('''\
+                    By default, the script will create a "build/rust" folder in the same folder as this
+                    script and build each requested stage within that containing folder. To change the
+                    location of the containing build folder, pass it to this parameter. This can be either
+                    an absolute or relative path. If it is provided, then a custom LLVM install folder
+                    needs to be provided as well to prevent mistakes.
+
+                    '''),
+                    type=str)
+parser.add_argument('-i',
+                    '--install-folder',
+                    help=textwrap.dedent('''\
+                    By default, the script will leave the toolchain in its build folder. To install it
+                    outside the build folder for persistent use, pass the installation location that you
+                    desire to this parameter. This can be either an absolute or relative path.
+
+                    '''),
+                    type=str)
+parser.add_argument('-l',
+                    '--llvm-install-folder',
+                    help=textwrap.dedent('''\
+                    By default, the script will try to use a built LLVM by './build-llvm.py'. To use
+                    another LLVM installation (perhaps from './build-llvm.py --install-folder'), pass
+                    it to this parameter.
+
+                    '''),
+                    type=str)
+parser.add_argument('-R',
+                    '--rust-folder',
+                    help=textwrap.dedent('''\
+                    By default, the script will clone the Rust project into the tc-build repo. If you have
+                    another Rust checkout that you would like to work out of, pass it to this parameter.
+                    This can either be an absolute or relative path. Implies '--no-update'. When this
+                    option is supplied, '--ref' and '--use-good-revision' do nothing, as the script does
+                    not manipulate a repository it does not own.
+
+                    '''),
+                    type=str)
+parser.add_argument('-n',
+                    '--no-update',
+                    help=textwrap.dedent('''\
+                    By default, the script always updates the Rust repo before building. This prevents
+                    that, which can be helpful during something like bisecting or manually managing the
+                    repo to pin it to a particular revision.
+
+                    '''),
+                    action='store_true')
+parser.add_argument('-r',
+                    '--ref',
+                    help=textwrap.dedent('''\
+                    By default, the script builds the main branch (tip of tree) of Rust. If you would
+                    like to build an older branch, use this parameter. This may be helpful in tracking
+                    down an older bug to properly bisect. This value is just passed along to 'git checkout'
+                    so it can be a branch name, tag name, or hash. This will have no effect if
+                    '--rust-folder' is provided, as the script does not manipulate a repository that it
+                    does not own.
+
+                    '''),
+                    default='master',
+                    type=str)
+parser.add_argument('--show-build-commands',
+                    help=textwrap.dedent('''\
+                    By default, the script only shows the output of the comands it is running. When this option
+                    is enabled, the invocations of the build tools will be shown to help with reproducing
+                    issues outside of the script.
+
+                    '''),
+                    action='store_true')
+clone_options.add_argument('--use-good-revision',
+                           help=textwrap.dedent('''\
+                    By default, the script updates Rust to the latest tip of tree revision, which may at times be
+                    broken or not work right. With this option, it will checkout a known good revision of Rust
+                    that builds and works properly. If you use this option often, please remember to update the
+                    script as the known good revision will change. This option may work best with a matching good
+                    revision used to build LLVM by './build-llvm.py'.
+
+                           '''),
+                           action='store_const',
+                           const=GOOD_REVISION,
+                           dest='ref')
+parser.add_argument('--vendor-string',
+                    help=textwrap.dedent('''\
+                    Add this value to the Rust version string (like "rustc ... (ClangBuiltLinux)"). Useful when
+                    reverting or applying patches on top of upstream Rust to differentiate a toolchain built
+                    with this script from upstream Rust or to distinguish a toolchain built with this script
+                    from the system's Rust. Defaults to ClangBuiltLinux, can be set to an empty string to
+                    override this and have no vendor in the version string.
+
+                    '''),
+                    type=str,
+                    default='ClangBuiltLinux')
+args = parser.parse_args()
+
+# Start tracking time that the script takes
+script_start = time.time()
+
+# Folder validation
+tc_build_folder = Path(__file__).resolve().parent
+src_folder = Path(tc_build_folder, 'src')
+
+if args.build_folder:
+    build_folder = Path(args.build_folder).resolve()
+
+    if not args.llvm_install_folder:
+        raise RuntimeError(
+            'Build folder customized, but no custom LLVM install folder provided -- this is likely a mistake. Provide both if you want to build in a custom folder?'
+        )
+else:
+    build_folder = Path(tc_build_folder, 'build/rust')
+
+if args.llvm_install_folder:
+    llvm_install_folder = Path(args.llvm_install_folder).resolve()
+else:
+    llvm_install_folder = Path(tc_build_folder, 'build/llvm/final')
+
+# Validate and configure Rust source
+if args.rust_folder:
+    if not (rust_folder := Path(args.rust_folder).resolve()).exists():
+        raise RuntimeError(f"Provided Rust folder ('{args.rust_folder}') does not exist?")
+else:
+    rust_folder = Path(src_folder, 'rust')
+rust_source = RustSourceManager(rust_folder)
+rust_source.download(args.ref)
+if not (args.rust_folder or args.no_update):
+    rust_source.update(args.ref)
+
+# Build Rust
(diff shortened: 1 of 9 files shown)


tc_build/rust.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def git_capture(self, cmd):
+        return self.git(cmd, capture_output=True).stdout.strip()
+
+    def ref_exists(self, ref):
+        try:
+            self.git(['show-branch', ref])
+        except subprocess.CalledProcessError:
+            return False
+        return True
```
nathanchance, line 152: We could introduce `GitSourceManager` in `lkt/source.py` that has these and `is_shallow()` from `lkt/llvm.py` so that both pieces of code could inherit them and deduplicate these. It may be possible to include `download()` and `update()` with a little refactoring but I am fine with throwing `NotImplementedError` in the base class and leaving it up to the subclasses to implement, even with a little more duplication.

nathanchance (commented):
I will review this more in depth either later this week or early next week but one initial comment from a brief read.

build-rust.py
```diff
@@ -0,0 +1,166 @@
...
+
+if args.build_folder:
+    build_folder = Path(args.build_folder).resolve()
+else:
+    build_folder = Path(tc_build_folder, 'build/rust')
+
+if args.llvm_install_folder:
+    llvm_install_folder = Path(args.llvm_install_folder).resolve()
+else:
+    llvm_install_folder = Path(tc_build_folder, 'build/llvm/final')
```
nathanchance, line 136: Hmmm, I wonder if we want to look for an `llvm` folder above next to `build_folder.parent` instead…

nathanchance (commented):
I took this for a spin today and I think this is pretty good shape for a v1 but I will try to do an actual hard review over the next couple of days if I am able to.

I did work on my idea I suggested earlier and pushed that just to see if it worked and looked cleaner and I think it does: https://github.com/nathanchance/tc-build/commits/rust-review

build-rust.py
nathanchance: Actually, that would not work because the user's specified build folder may not have the same structure at the tc-build default. It should probably be an error for the user to specify `--build-folder` without `--llvm-install-folder`, as we can assume if they customized the build folder for `build-rust.py`, they likely did it for `build-llvm.py` and the default will not work.

ojeda:
> I took this for a spin today and I think this is pretty good shape for a v1 but I will try to do an actual hard review over the next couple of days if I am able to.
> 
> I did work on my idea I suggested earlier and pushed that just to see if it worked and looked cleaner and I think it does: https://github.com/nathanchance/tc-build/commits/rust-review

Thanks! I agree it looks better (sorry, I should have replied above). Should I put your commit in this PR and apply the fixup?

nathanchance (approved):
> Thanks! I agree it looks better (sorry, I should have replied above). Should I put your commit in this PR and apply the fixup?

No worries! Yes, I think that would be good.

> I am not sure how far you would want to modify the LLVM side or the `do_llvm()` step, so I decided to send this with the failing CI so that you see it.

We should probably fix the CI configuration since it is custom tailored but should we also put an informative message in `build-rust.py` if these files cannot be found in the provided LLVM install folder?

Couple more minor comments but I think I am pretty happy with this, at least as an initial revision. If you felt like any of the “nice to have”s in the initial PR message were worth following up on in the future, feel free to open issues here for tracking.

tc_build/rust.py
ojeda: Resolving as you did it and I am putting the commit here (thanks!).

build-rust.py
ojeda: Sounds good -- added. I wonder if it should be a warning, but it does not matter much since it is not too much work for a user. I documented the behavior too.

tc_build/rust.py
```diff
@@ -0,0 +1,165 @@
...
+    def build(self):
+        if not self.folders.build:
+            raise RuntimeError('No build folder set for build()?')
+        if not Path(self.folders.source, 'bootstrap.toml').exists():
+            raise RuntimeError('No bootstrap.toml in source folder, run configure()?')
+
+        build_start = time.time()
+        base_x_cmd = ['./x.py']
+        # 'install' is used for simplicity.
+        self.run_cmd([*base_x_cmd, 'install'], cwd=self.folders.source)
```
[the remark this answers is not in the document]
ojeda: Done! I was trying to imitate the original script, but probably too much :)

tc_build/rust.py
```diff
@@ -0,0 +1,165 @@
...
+
+                    [install]
+                    prefix = "{install_folder}"
+                    sysconfdir = "etc"
+
+                    [rust]
+                    debug = {toml_boolean(self.debug)}
+                    codegen-tests = false
+
+                    [target.x86_64-unknown-linux-gnu]
```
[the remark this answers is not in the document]
ojeda: So I took a look, and I think it may be simpler if we instead use the `configure` script they provide, which takes care of figuring out the target section for us. More details on that on the changelog for v2 that I will put below.

ojeda:
v2:

The biggest change is that instead of writing a `bootstrap.toml` manually, I used their `configure` script instead:

  - It uses the `dist` profile, which is geared towards users and distributions that do not plan to do development, and seems the right thing to do to benefit from any new defaults there anyway.
  
  - It figures out the `target.<target>`s for us for the `llvm-config` setting, so that is simpler. I checked that, indeed, in an emulated arm64 host, it fills it with the expected one (note: I just configured to see that, I did not run the build, so proper testing of that is of course very welcome!).

  - It writes a `boostrap.toml` that includes a copy of the documentation for each tweaked value, which is nice.

  - The script fills `configure-args` too, which is also nice.

The `dist` profile already enables `extended` and disables `download-ci-llvm`, so I removed those for simplicity (we could also explicitly keep them to ensure that does not change).

Moreover, I reviewed the rest of the `dist` profile and it sets a few other bits differently -- I leave some notes here before I forget about them:

  - They enable `llvm-bitcode-linker` and `lld`, which are disabled by default in the main configuration file. They do not give a good reason, though. `lld` is not used if one provides a `llvm-config` (and one gets a warning); and apparently `llvm-bitcode-linker` is so far only used for running nvptx tests (I guess they enabled it to be able to run all tests).
  
  - They set `compression-profile` to `balanced`, while previously it would have been the default from the main configuration file, i.e. `fast`. There is also `best`. We don't do tarballs so far anyway.

  - They disable `download-rustc`, but that is already the default in the main configuration file anyway, which is why I was not setting it. (Not sure why they do it explicitly, though.)

  - They set `channel` to `auto-detect` explicitly, so that it is not `dev`. This, in turn, means `optimized-compiler-builtins` should be set to `true` already, which is what I was doing manually, so I removed it. It will also mean `omit-git-hash` is `false` now, which seems OK (I didn't check what happens in the case of a tarball, though). They also say `download-ci-llvm` depends on the channel, but the main configuration file does not say so (anyway, they set that one to `false` explicitly).

  - They set `*-stages` to `2`, i.e. the defaults, but that should not matter since we use the `install` target which its default was `2` already.

  - They disable the recently added `compiletest-use-stage0-libtest`, which sounds like the right thing to do.

  - They do not do anything with the `change-id`. I did not keep it in v2 (which can be done `'--set', 'change-id=ignore'`) to keep it simpler and avoid having a `--set` option (it is just a few lines of output). In fact, I wonder if we should track it in tc-build like the "known good revision".

A few other notes:

  - I used their `--llvm-root` option, instead of `--llvm-config`, since the former adds the `bin/llvm-config` part.

  - I also changed a couple bits to do the configuration also inside the build folder (by the way, perhaps we should now factor out the `install_folder`), so now less changes should happen in the source folder (`git clean -xdf` now only reports a `__pycache__` after a build).

  - In turn, that means we can manually skip setting the `build-dir` (which does not have a custom flag for it, i.e. we would need to use `'--set', f'build.build-dir={self.folders.build}',`). It does mean we introduce an extra folder (`build`), but that seems fine and in fact may be for the better (because it is the default and because it "separates" the install and build folders when they are the same).

Finally, I modified the LLVM side a bit to unbreak the CI. It turns out we will need some changes on the LLVM (please see the `TODO` commit):

  - To add a few distribution components (on top of the build tools ones and others already built by the CI options): `llvm-config`, `llvm-headers` and `llvm-libraries`.

    For the moment, I enabled them unconditionally, but we probably want to put them separately. Perhaps as a new flag, enabled by default, e.g. `--library`, intended to be used when LLVM will be used as a library by something else, like `rustc`. I also considered `--rust`/`--for-rust` or similar, but that probably requires figuring out the exact set of things to add (and changing the list into a set), i.e. in case someone does not enable the build tools.

  - To avoid using the thin archives (i.e. `CMAKE_CXX_ARCHIVE_*`). If we have the `--library` flag, that could disable them. For the moment, for testing, I commented it out.

nathanchance (commented):
I wanted to get this fully reviewed and flushed out today but I ran out of time so I am just posting what I have and coming back to it later because I should not have left you hanging for so long.

I was able to test this on AArch64 and it worked fine until I got an error due to an old copy of binutils in my path so I am rerunning it now.

I want to run some measurements on how much the default install size would change if we just enabled the things that `--library` would need and just did not give an option to turn it off.

nathanchance:
> Great to hear at least the arm64 is not completely broken.

Better than not completely broken now :)

```
$ build/rust/final/bin/rustc -vV
rustc 1.89.0-nightly (449c80178 2025-06-02) (ClangBuiltLinux)
binary: rustc
commit-hash: 449c801783ecef2aad3ae03d6c9e4ac007de7d4b
commit-date: 2025-06-02
host: aarch64-unknown-linux-gnu
release: 1.89.0-nightly
LLVM version: 21.0.0
```

> Thanks for checking if we could do `--library` unconditionally too. That would be simpler for everyone (i.e. including users), but I assumed you probably wanted to keep the tarballs as small as possible.

While I do use tc-build for the kernel.org toolchains, I use [a custom `--install-targets`](https://github.com/nathanchance/env/blob/5d4db230b92c14d329b0d6973e21d7b776d7d572/python/pgo-llvm-builder/build.py#L119-L130) so a change to the default distribution list would not affect those tarballs. Doing the space measurements, I do not think I can justify doing the same for the kernel.org toolchains though, which I think is fine but I am open to discussion on it.

```
$ diskus default
465.24 MB (465,244,160 bytes)

$ diskus additional-files
983.01 MB (983,011,328 bytes)
```

I think that it would be reasonable to add those to the distributions list by default, as it only adds about 300 build targets in my testing and I would expect people using this to have a reasonable amount of space for build and install artifacts when the space requirements for the source is already rather large.

We could make it easier on people doing local builds if we could utilize thin archives in certain circumstances since my measurements show significant savings but I understand it might be tricky to safely do.

```
$ diskus additional-files-thin-archives
645.21 MB (645,214,208 bytes)
```

Additionally, we might be missing some things in the distribution list, as I get the following error when I run `build-rust.py --llvm-install-folder` against a folder that was generated using `build-llvm.py --install-targets distribution`.

```
thread 'main' panicked at src/bootstrap/src/lib.rs:1798:24:
src.symlink_metadata() failed with No such file or directory (os error 2) ("src = .../bin/llvm-cov")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Panic was initiated from src/bootstrap/src/lib.rs:1798:24
```

This has probably not been noticed because we run just the `install` target by default, which will install everything, not just what was in the distribution. The distribution is only useful as a shortcut by default when just building the toolchain, not installing it for future use as well.

nathanchance:
@ojeda Alright, I think I got everything into a mergeable state locally, which I pushed to:

https://github.com/nathanchance/tc-build/commits/build-rust-py

I sought to address the concerns I had with `distribution` bloat above by introducing the concept of distribution profiles, which should allow users to customize the distribution to fit their needs (I will eventually allow users to completely customize this by getting out of the way when `LLVM_DISTRIBUTION_COMPONENTS` is already specified). I also included a few updates on top. The verbose configure change is somewhat personal so we could add an option to customize it. Let me know what you think (I can push a separate PR at some point if that would help look it over as well).

ojeda:
It will be definitely be helpful! :)

Thanks a lot Nathan -- I will send a message to the mailing lists and Zulip about it so that others are aware.

PR #324, opened 2026-03-07
build-rust.py: Introduce '--configure-set-args'

nathanchance:
To give end users more flexibility in configuring their toolchain. This mirrors the `build-llvm.py` `--defines` option.

For example, this will allow me to build Rust using my custom environment for the kernel.org toolchains:

https://github.com/nathanchance/env/blob/8997cea59354fbe02e117100148ea48bd6ee88b3/python/pgo-llvm-builder/build.py#L411-L422

### DIFF
--- a/build-rust.py
+++ b/build-rust.py
@@ -35,6 +35,23 @@ parser.add_argument('-b',
 
                     '''),
                     type=str)
+parser.add_argument('-c',
+                    '--configure-set-args',
+                    help=textwrap.dedent('''\
+                    Pass the provided values to configure via the '--set' argument.
+
+                    For example:
+
+                        --configure-set-args build.foo=false build.bar=true
+
+                    would pass
+
+                        --set build.foo=false --set build.bar=true
+
+                    to configure.
+
+                    '''),
+                    nargs='+')
 parser.add_argument('-i',
                     '--install-folder',
                     help=textwrap.dedent('''\
@@ -161,6 +178,8 @@ final.folders.source = rust_folder
 final.folders.build = Path(build_folder, 'final')
 final.folders.install = Path(args.install_folder).resolve() if args.install_folder else None
 final.llvm_install_folder = llvm_install_folder
+if args.configure_set_args:
+    final.configure_set_args += args.configure_set_args
 final.debug = args.debug
 final.vendor_string = args.vendor_string
 final.show_commands = args.show_build_commands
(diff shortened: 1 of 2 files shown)


msfjarvis (approved):

ojeda:
> For example, this will allow me to build Rust using my custom environment for the kernel.org toolchains:

Nice to hear that is coming!

By the way, on your kernel.org changes, I saw:

```
                # We want to use our copy of LLVM to ensure full bitcode
                # compatibility
                'llvm.download-ci-llvm': 'false',
```

Did you see it downloading LLVM? From what I recall from https://github.com/ClangBuiltLinux/tc-build/pull/299#issuecomment-2856855272, that should be already disabled, i.e. I had it explicitly set in v1 of that PR, and then removed it when I moved to the `configure` script.

If you did it to ensure it (which I think it never hurts to be explicit here, in case they change something upstream -- after all, `build-rust.py`'s main use case is to match LLVM), then I think we should perhaps do it on `build-rust.py` by default?

Or perhaps we could query the values and check they are what we expect (if that is possible), which would allow us to detect changes upstream, and we could also do it for `gcc.download-ci-gcc` and `rust.download-rustc` too.

Thanks!

nathanchance:
> > For example, this will allow me to build Rust using my custom environment for the kernel.org toolchains:
> 
> Nice to hear that is coming!
> 
> By the way, on your kernel.org changes, I saw:
> 
> ```
>                 # We want to use our copy of LLVM to ensure full bitcode
>                 # compatibility
>                 'llvm.download-ci-llvm': 'false',
> ```
> 
> Did you see it downloading LLVM? From what I recall from [#299 (comment)](https://github.com/ClangBuiltLinux/tc-build/pull/299#issuecomment-2856855272), that should be already disabled, i.e. I had it explicitly set in v1 of that PR, and then removed it when I moved to the `configure` script.

At some point, I got an error due to a mismatch between the bitcode emitted by the kernel.org libLLVM (22.1.0) and Rust's (21.1.8-rust-1.93.0 or something like that) and I had figured it could be due to that. That said, I just tested without that and everything worked fine, so it is possible that one of the other things I did like using the same vanilla LLVM source via a symlink to `src/llvm-project` may have also helped avoid that, so I will drop that option. Thanks for making me double check!

ojeda:
I see, you're welcome! It could still be a good idea to enforce it perhaps, I am not sure (I was trying to minimize the "custom" `set`s we used in `./configure`).

PR #330, opened 2026-04-16
tc-build: Updates to ruff.toml (April 16, 2026)

nathanchance:
See the individual changes for the full details.

### DIFF
--- a/tc_build/tools.py
+++ b/tc_build/tools.py
@@ -1,11 +1,12 @@
 # pylint: disable=invalid-name
 
+from __future__ import annotations
+
 import os
 import re
 import shutil
 import subprocess
 from pathlib import Path
-from typing import Union
 
 import tc_build.utils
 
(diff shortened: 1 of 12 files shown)


tc_build/tools.py
```diff
@@ -11,6 +11,22 @@
...
+def generate_versioned_binaries() -> list[str]:
+    try:
+        cmakelists_txt = tc_build.utils.curl(
+            'https://raw.githubusercontent.com/llvm/llvm-project/main/cmake/Modules/LLVMVersion.cmake'
+        )
+    except subprocess.CalledProcessError:
+        llvm_tot_ver = 23
+    else:
+        if not (match := re.search(r'set\(LLVM_VERSION_MAJOR\s+(\d+)', cmakelists_txt)):
+            msg = 'Could not find LLVM_VERSION_MAJOR in CMakeLists.txt?'
```
msfjarvis, line 23: Non-blocking: This is not searching CMakeLists.txt, might be worth changing?

msfjarvis (approved):

tc_build/tools.py
nathanchance: Great catch: #331

PR #346, opened 2026-08-08
tc_build: source: Only fetch the requested ref when cloning shallow

danascape:
'--shallow-clone' with a ref other than main adds '--no-single-branch' to work around '--depth=1' implying '--single-branch', so that the checkout after the clone succeeds. That fetches the tip of every ref, which is thousands of them for llvm-project, making a shallow clone nearly as slow as a full one.

Use '--branch' instead, which fetches only the requested ref and works for both branches and tags. Hashes were already unsupported here, so nothing is lost.

### DIFF
--- a/build-llvm.py
+++ b/build-llvm.py
@@ -448,8 +448,8 @@ clone_options.add_argument(
 
                     1. This cannot be used with '--use-good-revision'.
 
-                    2. When no '--branch' is specified, only main is fetched. To work with other branches,
-                       a branch other than main needs to be specified when the repo is first cloned.
+                    2. Only the ref specified by '--ref' (main by default) is fetched. To work with a
+                       different ref, the repo has to be cloned again with that ref.
 
                            '''),
     action='store_true',
--- a/tc_build/source.py
+++ b/tc_build/source.py
@@ -112,7 +112,7 @@ class GitSourceManager:
         if shallow:
             git_clone.append('--depth=1')
             if ref != 'main':
-                git_clone.append('--no-single-branch')
+                git_clone.append(f"--branch={ref}")
         git_clone += [self._repo_url, self.repo]
 
         subprocess.run(git_clone, check=True)


nathanchance:
Thanks for the patch! Truth be told though, I think we should just kill `--shallow-clone`. It was always a somewhat haphazard operation and now it definitely does not work with the `users/` branches, as you mention.

If people want to shallow clone the repo, they are free to just do it manually themselves and invoke the script with `--llvm-folder`.

danascape:
> Thanks for the patch! Truth be told though, I think we should just kill `--shallow-clone`. It was always a somewhat haphazard operation and now it definitely does not work with the `users/` branches, as you mention.
> 
> If people want to shallow clone the repo, they are free to just do it manually themselves and invoke the script with `--llvm-folder`.

yep, I manually cloned and passed --llvm-folder to compile, but wouldn't that beat the purpose of shallow-clone itself, the place I live at, I cannot survive the time it will take for a full-clone, hence did that patch and one-line command

nathanchance:
> yep, I manually cloned and passed --llvm-folder to compile, but wouldn't that beat the purpose of shallow-clone itself, the place I live at, I cannot survive the time it will take for a full-clone, hence did that patch and one-line command

Not necessarily, it would just require the user to do the shallow clone before running the script.

That said, the more I think about it, I believe this is fine. `--no-single-branch` was added in d0abcf1145cb904decd6cc31b153ef5bfef42279 to fix #81 but now that we detect a shallow repository with that problematic situation (2b95696e1cc8f9f778e40be7afae9ab1b915ef91, survives in [`tc_build/source.py`](https://github.com/ClangBuiltLinux/tc-build/blob/a8d972659da5772fc6ebb73cd5518cabca3f3b83/tc_build/source.py#L148)), it should be fine to restrict the clone to that single branch. We still discourage shallow cloning. I do have some ideas about how to save some size on the clone, namely using `--filter=blob:none` and using a negative refspec for `users/*` and `revert-*` [as suggested in LLVM's documentation](https://llvm.org/docs/GettingStarted.html#getting-the-source-code-and-building-llvm) but I want to make sure they work well before making such a change.

nathanchance (approved):

